PR #6, opened 2020-12-16
Enable workflow from pr

siyuan0322:
## What do these changes do?
Removes the needs of permission push to registry from PR workflow.

1. Use artifact instead of registry push-and-pull to transfer image between jobs.
2. Remove unused environment variables
3. Reduce the time in Build Image stage by not build unnecessary executables.
4. Reduce logs that are too verbose.
5. Add cache for pip packages.


sighingnow (approved):
LGTM 👍

PR #36, opened 2020-12-23
Support bytecode-based requests for interactive engine

### LINKED ISSUE #35: Support bytecode-based requests in interactive engine.
Interactive engine already supports submit a script to query the remote graph, which syntax is like:
`interactive.execute("g.V().count()").result().one()`.

However, there is also another bytecode-based style to query a graph, which should also be supported, as follows:
`g.V().count().toList()`.

We should support bytecode-based style as it's the recommended style by Tinkerpop.

siyuan0322:
## What do these changes do?
Support bytecode-based requests for interactive engine.
User can get a `GraphTraversalSource` to traversal a remote graph by using `interactive.traversal_source()` where `interactive` is created by `sess.gremlin(graph)`.

Examples:
```
>>> sess = graphscope.session()
>>> graph = load_modern_graph(sess, modern_graph_data_dir)
>>> interactive = sess.gremlin(graph)
>>> g = interactive.traversal_source()
>>> g.V().both()[1:3].toList()
[v[0.1], v[0.2]]
>>> g.V().both().name.toList()
['vadas', 'peter', 'lop', 'josh', 'marko', 'ripple']
```
## Related issue number

Fixes #35

### DIFF
--- a/python/graphscope/interactive/query.py
+++ b/python/graphscope/interactive/query.py
@@ -42,7 +42,7 @@ class InteractiveQuery(object):
     which lifetime is independent from the origin graph.
 
     Use `execute()` to submit a script,
-    ans use `traversal_source()` to get a `GraphTraversalSource` for further traversal.
+    and use `traversal_source()` to get a `GraphTraversalSource` for further traversal.
     """
 
     def __init__(self, graphscope_session, object_id, front_ip, front_port):
--- a/python/graphscope/interactive/query.py
+++ b/python/graphscope/interactive/query.py
@@ -41,8 +41,8 @@ class InteractiveQuery(object):
     from origin graph, produce a new, smaller but concise graph stored in vineyard,
     which lifetime is independent from the origin graph.
 
-    Use `execute()` to submit a script,
-    and use `traversal_source()` to get a `GraphTraversalSource` for further traversal.
+    User can either use `execute()` to submit a script, or use `traversal_source()`
+    to get a `GraphTraversalSource` for further traversal.
     """
 
     def __init__(self, graphscope_session, object_id, front_ip, front_port):
--- a/python/graphscope/interactive/query.py
+++ b/python/graphscope/interactive/query.py
@@ -149,12 +149,12 @@ class InteractiveQuery(object):
 
     def traversal_source(self):
         """Create a GraphTraversalSource and return.
-        Once "g" has been created using a connection, we can start to write
+        Once `g` has been created using a connection, we can start to write
         Gremlin traversals to query the remote graph.
 
         Examples:
 
-            ..code:: python
+            .. code:: python
 
                 sess = graphscope.session()
                 graph = load_modern_graph(sess, modern_graph_data_dir)
--- a/python/graphscope/interactive/query.py
+++ b/python/graphscope/interactive/query.py
@@ -22,8 +22,8 @@ import random
 from concurrent.futures import ThreadPoolExecutor
 
 from gremlin_python.driver.client import Client
-from gremlin_python.process.anonymous_traversal import traversal
 from gremlin_python.driver.driver_remote_connection import DriverRemoteConnection
+from gremlin_python.process.anonymous_traversal import traversal
 
 from graphscope.framework.loader import Loader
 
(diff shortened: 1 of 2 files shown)


siyuan0322:
/CC @yecol for a review,
and /CC @zhengpingq since it's an interactive engine related PR.

python/graphscope/interactive/query.py
```diff
@@ -32,11 +34,15 @@ class InteractiveQuery(object):
...
-    It also can expose gremlin endpoint which can be used by any other standard gremlin console.
+    It also can expose gremlin endpoint which can be used by
+    any other standard gremlin console, with the method `graph_url()`.
 
     It also has a method called `subgraph` which can extract some fragments
     from origin graph, produce a new, smaller but concise graph stored in vineyard,
     which lifetime is independent from the origin graph.
+
+    Use `execute()` to submit a script,
+    ans use `traversal_source()` to get a `GraphTraversalSource` for further traversal.
```
sighingnow, line 45: `ans` -> `and`.

Could you please use a more readable line-breaks in comment? The following would be better:

```
Use `execute()` to submit a script, and use `traversal_source()` to get a `GraphTraversalSource`
for further traversal.
```

That doesn't affect the generated docs but will benifit the readability of source code. It would be great to think you are writing a document or paper when writing docstring in comments.

python/graphscope/interactive/query.py
```diff
@@ -141,6 +147,27 @@ def execute(self, query):
...
         return self._client.submit(query)
 
+    def traversal_source(self):
+        """Create a GraphTraversalSource and return.
+        Once "g" has been created using a connection, we can start to write
+        Gremlin traversals to query the remote graph.
+
+        Examples:
+
+            ..code:: python
```
sighingnow, line 157: Bad indent.

```
Examples:

.. code:: python

    xxxxx
```

python/graphscope/interactive/query.py
```diff
@@ -141,6 +147,27 @@ def execute(self, query):
...
+                graph = load_modern_graph(sess, modern_graph_data_dir)
+                interactive = sess.gremlin(graph)
+                g = interactive.traversal_source()
+                print(g.V().both()[1:3].toList())
+                print(g.V().both().name.toList())
+
+        Returns:
+            `GraphTraversalSource`
+        """
+        return traversal().withRemote(DriverRemoteConnection(self._graph_url, "g"))
```
sighingnow, line 169: Is the parameter `"g"` a hard-coded constant value?

sighingnow (changes requested):
/cc @shirly121

python/graphscope/interactive/query.py
siyuan0322: AFAIK yes.

python/graphscope/interactive/query.py
siyuan0322: I think sections are supposed to indent after the section header?
ref: https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html#example-google

python/graphscope/interactive/query.py
sighingnow: Then could `"g"` will be used repeatly?

python/graphscope/interactive/query.py
siyuan0322: Yes, I already noticed the typo and have a fix locally, and I'm thinking combine the typo fixes with other comments to avoid another unnecessary CI.

python/graphscope/interactive/query.py
sighingnow: I didn't realize the `Examples:` is a directive here, but please double check the generated doc.

python/graphscope/interactive/query.py
siyuan0322: Yes if you are talking about using `g` repeatedly across graph instances.

sighingnow: What will happen on the same graph instance?

siyuan0322: Nothing happens, you can get two traversal source that works normally.
They are same if you look at the initialization of `Client` in https://github.com/alibaba/GraphScope/blob/dd2a58ac8ddd1600a997ef0312895e5f87768168/python/graphscope/interactive/query.py#L46

sighingnow (approved):

PR #49, opened 2020-12-25
Fixed the version at the time of releasing the pip package.

### LINKED ISSUE #26: [BUG] Don't used `:latest` tag in python as default image

lidongze0629:
## What do these changes do?
- Add `__version__` in graphscope module.
```python
import graphscope
graphscope.__version__
```
- ~~Fix the version at the time of releasing the pip package.~~

## Related issue number

Fixes #26


python/setup.py
```diff
@@ -54,6 +54,77 @@ def run(self):
...
+
+    def finalize_options(self):
+        pass
+
+    def run(self):
+        from graphscope._version import __version__
+        from graphscope._version import git_info
+
+        tag = __version__
+        commit_hash, commit_ref = git_info()
```
siyuan0322, line 77: As there are many plain `return` in `git_info`, which means if there are any errors in `git_info`, you will got a `None` as return value, which will trigger a `TypeError: cannot unpack non-iterable NoneType object` in `commit_hash, commit_ref = git_info()`.

python/graphscope/_version.py
```diff
@@ -25,5 +25,5 @@
 try:
     with open(os.path.join(git_root, "VERSION")) as f:
         __version__ = f.readline().rstrip()
-except:
+except:  # noqa: E722
```
siyuan0322, line 28: Why do we need the broad except...  I think there might a FileNotFoundError.  or OSError. anything else?

scripts/prepare_env.sh
```diff
@@ -6,6 +6,9 @@ set -e
 set -x
 set -o pipefail
 
+graphscope_home="$( cd "$(dirname "$0")/.." >/dev/null 2>&1 ; pwd -P )"
+__version__=`cat ${graphscope_home}/VERSION`
```
siyuan0322, line 10: Weird... I think using `version=xxx` is ok.. why do we need a `__version__`.
and use $(command) instead of `command`

VERSION
```diff
@@ -0,0 +1 @@
+v0.1.0
```
sighingnow, line 1: No `v` prefix in version number.

coordinator/gscoordinator/coordinator.py
```diff
@@ -43,6 +43,7 @@
 sys.stdout = StdoutWrapper(sys.stdout)
 
 from daemons.prefab import run
```
sighingnow, line 45: Seems the `daemons` isn't used anymore?

python/graphscope/__init__.py
```diff
@@ -16,6 +16,7 @@
 # limitations under the License.
 #
 
+from graphscope._version import __version__
```
sighingnow, line 19: No need for `_` in `graphscope._version` module name.

sighingnow (changes requested):
Note that there are version number in `setup.py` in `python/` and `coordinator` for graphscope and gs-coordinator packages.

python/graphscope/_version.py
lidongze0629: Yes, Here we want is cover everything exceptions occur and then set the version `latest`.  Change to `OSError `.

sighingnow (approved):
LGTM.

sighingnow:
Please review the "coordinator/setup.py" as "python/setup.py".

PR #71, opened 2021-01-07
Upgrade vineyard to v0.1.5.

sighingnow:
## What do these changes do?

Bump up vineyard version, v0.1.5 import some incompatible changes between new client and old server.

## Related issue number

N/A


sighingnow:
Close as infavour of #73 to make CI happy.

PR #101, opened 2021-01-24
enable  create session from an endpoint

acezen:
enable  create session from an endpoint


acezen:
cc/ @sighingnow @lidongze0629

sighingnow (approved):
LGTM.

siyuan0322 (approved):
LGTM

coordinator/gscoordinator/coordinator.py
```diff
@@ -146,10 +146,11 @@ def __init__(self, launcher, dangling_seconds, log_level="INFO"):
 
         # dangling check
         self._dangling_seconds = dangling_seconds
-        self._dangling_detecting_timer = threading.Timer(
-            interval=self._dangling_seconds, function=self._cleanup, args=(True,)
-        )
-        self._dangling_detecting_timer.start()
+        if self._dangling_seconds != -1:
```
lidongze0629, line 149: Would it be better?
```python
if self._dangling_seconds >= 0:
    xxxx
```

python/graphscope/client/session.py
```diff
@@ -273,7 +273,7 @@ def __init__(
         # update other optional params
         self._config_params.update(kw)
 
-        self._config_params["addr"] = None
+        self._config_params["addr"] = kw.pop("addr", None)
```
lidongze0629, line 276: Maybe you can update the session doc if we support param `addr`.

lidongze0629 (approved):

coordinator/gscoordinator/coordinator.py
acezen: good suggestion, fixed.

PR #103, opened 2021-01-25
Add serialize and deserialize method for graph.

siyuan0322:


yecol:
Please verify the correctness (or return error messages) for edge cases, e.g., 
- serialized on 2 workers, deserialize from 3 workers.
- serialized on 4 workers, deserialize from another 4 workers, but they sharing the same worker_0
- serialization files are not complete, e.g., in one of the workers, the blob file is missing.
- ...

sighingnow (approved):

PR #115, opened 2021-01-29
Requires vineyard-io, and drop the copy hack.

sighingnow:
That is introduced in https://github.com/alibaba/GraphScope/pull/103.


siyuan0322 (approved):
👍

sighingnow:
Merged since the occasional failure is not caused by this changes.

Issue #114, opened 2021-01-28
Deprecate the codepath for minikube (if kind workers better for Mac) to simply our environment prepration steps.


sighingnow:
**Is your feature request related to a problem? Please describe.**

Currently the "Prerequisites" section in README is a bit confusion, e.g., do we advertise that kind is OK as well for Mac? AFAIK we haven't test it yet.

And the codepath for minikube to get the service endpoint on Mac, as well as the `k8s_minikube_vm_driver` parameter in `graphscope.session` is a bit hacky.

**Describe the solution you'd like**

If kind works better we could deprecate the support for minikube and use kind everywhere in our codebase and documentation.

**Additional context**

N/A

acezen:
the 'kind' is not very OK for Mac, it need some hack tool like [docker-tuntap-osx](https://github.com/AlmirKadric-Published/docker-tuntap-osx) and [MetalLB](https://metallb.universe.tf/) to  Load Balancing on Mac.

Issue #106, opened 2021-01-25
[BUG] Figure out the required protobuf version for the graphscope python package.


sighingnow:
**Describe the bug**

GraphScope will raise if the existing protobuf package version is too low:

```
AttributeError: module 'google.protobuf.descriptor' has no attribute '_internal_create_key'
```

That is because we didn't specify the version range of protobuf in `requirements.txt`. That is bad when raising during `import graphscope`, and without a clear error message to tell the users how to fix that.

**Expected behavior**

Declare which version that we could accept in `requirements.txt`.

**Additional context**

N/A.

lidongze0629:
protobuf>=v3.12.0

PR #129, opened 2021-02-05
Deprecated 'k8s_minikube_vm_driver' parameter in session.

### LINKED ISSUE #54: Remove all `show_logs=True` parameter when initializing a session.[BUG]
**Describe the bug**

The `show_log=True` parameter has been removed in #42 but there are still many occurrences in our codebase.

**Expected behavior**

Don't use removed parameter.

lidongze0629:
## What do these changes do?

Minikube can not be deployed with non-driver in mac OS,  so we deprecate the k8s_minikube_vm_driver in session which is a bit hacky.
This PR also fixes #54 #106 

## Related issue number

Fixes #54 #106 #114

### DIFF
--- a/python/tests/test_session.py
+++ b/python/tests/test_session.py
@@ -29,6 +29,8 @@ import pytest
 import graphscope
 from graphscope.client.session import DEFAULT_CONFIG_FILE
 
+graphscope.set_option(show_log=True)
+
 COORDINATOR_HOME = os.path.join(os.path.dirname(__file__), "../", "../coordinator")
 new_data_dir = os.path.expandvars("${GS_TEST_DIR}/new_property/v2_e2")
 
@@ -120,7 +122,7 @@ def load_graph(session):
 
 
 def test_default_session():
-    s = graphscope.session(run_on_local=True, show_log=True)
+    s = graphscope.session(run_on_local=True)
 
     info = s.info
     assert info["status"] == "active"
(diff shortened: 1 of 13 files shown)


python/tests/test_session.py
```diff
@@ -29,6 +29,8 @@
 import graphscope
 from graphscope.client.session import DEFAULT_CONFIG_FILE
 
+graphscope.set_option(show_log=True)
```
sighingnow, line 32: Put the global setting to `setUpModule`.

c.f.: https://docs.python.org/3/library/unittest.html#setupmodule-and-teardownmodule

sighingnow (approved):
I have noticed comment from Weibin in #114:

> the 'kind' is not very OK for Mac, it need some hack tool like [docker-tuntap-osx](https://github.com/AlmirKadric-Published/docker-tuntap-osx) and [MetalLB](https://metallb.universe.tf/) to Load Balancing on Mac.
> 
> I think the Kubernetes of Docker for Mac with Load Balancing is enough to getting started on Mac.
> 
> ref: https://www.thehumblelab.com/kind-and-metallb-on-mac/#_

Could you please drop all word "Minikube" in documentation (as well as README.md)? It there's any...

acezen:
Do we need to replace minikube with KinD since the prepare_env.sh use KinD on Linux/WSL2?

sighingnow:
> Do we need to replace minikube with KinD since the prepare_env.sh use KinD on Linux/WSL2?

Continuing to use minikube in CI is fine for me, that would be helpful for avoiding further debugging work for CI. The usage of minikube is transparent, though. (and won't affect end users)

PR #152, opened 2021-02-20
optimize source chain operator

siyuan0322:
Description
-----------

GIE scan all vertices unnecessarily for a simple query like `g.V().has('label','id', xxx).out()`.
It should use primary key to find the exact vertex.

Relational Context
------------------

This bug was found in an experiment of an industrial paper, where some simple query cost too much time to execute.


sighingnow (approved):
LGTM 👍

PR #162, opened 2021-02-24
Capture KeyboardInterrupt during the session launching and cleanup re…

### LINKED ISSUE #160: [BUG] Session still running on the backend when the ctrl-c signal received
**Describe the bug**
It is still running on the backend when the ctrl-c signal received during the process of session launching.

**To Reproduce**
1. ipython
2. sess = graphscope.session()
3. ctrl-c

**Expected behavior**
Handle the `KeyboardInterrupt`, stop and destruct the session.

lidongze0629:
## What do these changes do?

Fixes #160 by capture and handle the KeyboardInterrupt during the session launching.

## Related issue number

Fixes #160

### DIFF
--- a/python/graphscope/client/rpc.py
+++ b/python/graphscope/client/rpc.py
@@ -96,6 +96,7 @@ class GRPCClient(object):
     def waiting_service_ready(self, timeout_seconds=60, enable_k8s=True):
         begin_time = time.time()
         request = message_pb2.HeartBeatRequest()
+        response = None
         while True:
             try:
                 response = self._stub.HeartBeat(request)
--- a/python/graphscope/client/rpc.py
+++ b/python/graphscope/client/rpc.py
@@ -96,6 +96,7 @@ class GRPCClient(object):
     def waiting_service_ready(self, timeout_seconds=60, enable_k8s=True):
         begin_time = time.time()
         request = message_pb2.HeartBeatRequest()
+        # Do not drop this line, which is for handling KeyboardInterrupt.
         response = None
         while True:
             try:
--- a/python/graphscope/client/utils.py
+++ b/python/graphscope/client/utils.py
@@ -87,27 +87,28 @@ class CaptureKeyboardInterrupt(object):
     """Context Manager for capture keyboard interrupt
 
     Args:
-        func: function
+        callback: function
             Callback function when KeyboardInterrupt occurs.
 
     Examples:
-        >>> with CaptureKeyboardInterrupt(func):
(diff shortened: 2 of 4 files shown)


python/graphscope/client/rpc.py
```diff
@@ -96,6 +96,7 @@ def __init__(self, endpoint):
     def waiting_service_ready(self, timeout_seconds=60, enable_k8s=True):
         begin_time = time.time()
         request = message_pb2.HeartBeatRequest()
+        response = None
```
sighingnow, line 99: No need for that.

python/graphscope/client/utils.py
```diff
@@ -83,6 +83,33 @@ def update():
 GSLogger.init()
 
 
+class CaptureKeyboardInterrupt(object):
+    """Context Manager for capture keyboard interrupt
+
+    Args:
+        func: function
```
sighingnow, line 90: naming: `func` -> `callback` or `deleter`.

python/graphscope/client/utils.py
```diff
@@ -83,6 +83,33 @@ def update():
...
+    """
+
+    def __init__(self, func=None):
+        self._func = func
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, exc_tb):
+        if exc_type is KeyboardInterrupt:
```
sighingnow, line 105: Otherwise we don't need such cleanup?

I remember that if the `connect` throws exceptions the kubernetes namespace won't be correctly (though I doesn't remember it clearly).

sighingnow (approved):
LGTM.

python/graphscope/client/rpc.py
lidongze0629: This is for line-107 in finally statement: `if response is not None`.

When Ctrl-C signal occurs, The try statement cannot guarantee that `response` is always defined.

sighingnow: Thanks for clearfication. So leave a comment hear please to prevent this line being drop by accident: `KeyboardInterrupt` is not an `Exception`, rather, it inherits from `BaseException`.

python/graphscope/client/utils.py
lidongze0629: Great suggestion. I have captured all exceptions here.

python/graphscope/client/utils.py
lidongze0629: done

Issue #157, opened 2021-02-23
Improve the response time of each step in Python


yecol:
Currently, some steps are response in a long time in Python, due to k8s pods/container launching, app/graph compilation... 
To reduce the response time, we may need to profile the steps, break down each part of cost in the engines, and find out the bottleneck.

lidongze0629:
Related Issues:

#137 Pre-compile common apps/graphs in the distributed docker image

yecol:
Hi Dongze, besides issue #137, you may also want to profiling other operations, e.g., `.gremlin()`, `.draw()`, ...

lidongze0629:
**Rough profiling for each step:**

launch session:  29s   (image pulled already)
load graph: 54s (compile time: 40s, load time: 14s)
sess.gremlin:   50s
interactive subgraph time: 19s (load graph)
project_to_simple time: 26s
build_app time: 26s
add_column time: 3s

lidongze0629:
**Reopen this issue to record the latest time(after improved) for each step.**

launch session:  
load graph: 
sess.gremlin:   
interactive subgraph time: 
project_to_simple time: 
build_app time: 
add_column time:
draw time: 
close time: 

**Some case for further optimization**

PR #170, opened 2021-03-01
Optimize time of sess.gremlin and sess.close

lidongze0629:
## What do these changes do?

- Deleted several unnecessary `sleep` operations during process of gremlin launching
- Launch gremlin server's pod  implicitly when a property graph created.

## Related issue number

Fixes https://github.com/alibaba/GraphScope/issues/157

### DIFF
--- a/python/graphscope/client/session.py
+++ b/python/graphscope/client/session.py
@@ -57,6 +57,8 @@ from graphscope.framework.errors import K8sError
 from graphscope.framework.errors import LearningEngineInternalError
 from graphscope.framework.errors import check_argument
 from graphscope.framework.operation import Operation
+from graphscope.interactive.query import InteractiveQuery
+from graphscope.interactive.query import InteractiveQueryStatus
 from graphscope.proto import message_pb2
 from graphscope.proto import op_def_pb2
 from graphscope.proto import types_pb2
@@ -894,15 +896,29 @@ class Session(object):
         handle_json_string = json.dumps(handle)
         return base64.b64encode(handle_json_string.encode("utf-8")).decode("utf-8")
 
-    def gremlin(self, graph, engine_params=None, timeout_seconds=60):
+    @set_defaults(gs_config)
+    def gremlin(self, graph, engine_params=None):
         """Get a interactive engine handler to execute gremlin queries.
+
+        Note that this method will be executed implicitly when a property graph created
+        and cache a instance of InteractiveQuery in session if `initializing_interactive_engine`
+        is True. If you want to create a new instance under the same graph by different params,
+        you should close the instance first.
+
+        .. code:: python
+
+            >>> # close and recreate InteractiveQuery.
+            >>> interactive_query = sess.gremlin(g)
+            >>> interactive_query.close()
+            >>> interactive_query = sess.gremlin(g, engine_params={"xxx":"xxx"})
+
 
         Args:
             graph (:class:`Graph`): Use the graph to create interactive instance.
             engine_params (dict, optional): Configure startup parameters of interactive engine.
+                You can also configure this param by `graphscope.set_option(engine_params={})`.
                 See a list of configurable keys in
                 `interactive_engine/deploy/docker/dockerfile/executor.vineyard.properties`
-            timeout_seconds (int, optional): For waiting gremlin service ready. Defaults to 60.
 
         Raises:
             InvalidArgumentError: :code:`graph` is not a property graph or unloaded.
@@ -911,32 +927,35 @@ class Session(object):
             :class:`InteractiveQuery`
         """
 
+        # self._interactive_instance_dict[graph.vineyard_id] will be None if
+        # InteractiveQuery closed
         if (
             graph.vineyard_id in self._interactive_instance_dict
             and self._interactive_instance_dict[graph.vineyard_id] is not None
         ):
             interactive_query = self._interactive_instance_dict[graph.vineyard_id]
-            if interactive_query.status == "running":
+            if interactive_query.status == InteractiveQueryStatus.Running:
                 return interactive_query
-            elif interactive_query.status == "pending":
-                begin_time = time.time()
+            elif interactive_query.status == InteractiveQueryStatus.Failed:
+                raise InteractiveEngineInternalError(interactive_query.error_msg)
+            else:
+                # Initializing.
+                # while True is ok, as the status is either running or failed eventually after timeout.
                 while True:
                     time.sleep(1)
-                    if interactive_query.status == "running":
+                    if interactive_query.status == InteractiveQueryStatus.Running:
                         return interactive_query
-                    if time.time() - begin_time >= timeout_seconds:
-                        raise TimeoutError("Launch gremlin server timeout")
+                    elif interactive_query.status == InteractiveQueryStatus.Failed:
+                        raise InteractiveEngineInternalError(
+                            interactive_query.error_msg
+                        )
 
         if not graph.loaded():
             raise InvalidArgumentError("The graph has already been unloaded")
         if not graph.graph_type == types_pb2.ARROW_PROPERTY:
             raise InvalidArgumentError("The graph should be a property graph.")
 
-        from graphscope.interactive.query import InteractiveQuery
-
-        interactive_query = InteractiveQuery(
-            graphscope_session=self, object_id=graph.vineyard_id
-        )
+        interactive_query = InteractiveQuery(session=self, object_id=graph.vineyard_id)
         self._interactive_instance_dict[graph.vineyard_id] = interactive_query
 
         if engine_params is not None:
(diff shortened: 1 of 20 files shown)


python/graphscope/framework/graph.py
```diff
@@ -156,6 +164,14 @@ def _close_learning_instances(self):
             instance.close()
         self._learning_instance_list.clear()
 
+    def _launch_interactive_instance_impl(self):
+        try:
+            sess = get_session_by_id(self.session_id)
+            sess.gremlin(self)
+        except:  # noqa: E722
+            # suppress all exceptions
+            pass
```
sighingnow, line 173: Record the exception, and when failed, rethrow when users call `.gremlin()`.

python/graphscope/interactive/__init__.py
```diff
@@ -18,8 +18,12 @@
...
 # jupyter notebook's event loop.
-import nest_asyncio
 
-nest_asyncio.apply()
+try:
+    import nest_asyncio
+
+    nest_asyncio.apply()
+except:  # noqa: E722
+    pass
```
sighingnow, line 27: Why skip? I remember that we have already installed `nest_asyncio` into our dependencies in `requirements.txt`.

python/graphscope/interactive/query.py
```diff
@@ -66,9 +73,21 @@ def graph_url(self):
         """The gremlin graph url can be used with any standard gremlin console, e.g., thinkerpop."""
         return self._graph_url
 
+    @property
+    def status(self):
+        return self._status
```
sighingnow, line 78: Define an enum for status.

sighingnow (approved):

python/graphscope/client/session.py
```diff
@@ -913,20 +918,36 @@ def gremlin(self, graph, engine_params=None):
...
             raise InvalidArgumentError("The graph has already been unloaded")
         if not graph.graph_type == types_pb2.ARROW_PROPERTY:
             raise InvalidArgumentError("The graph should be a property graph.")
 
+        from graphscope.interactive.query import InteractiveQuery
+
+        interactive_query = InteractiveQuery(
+            graphscope_session=self, object_id=graph.vineyard_id
+        )
+        self._interactive_instance_dict[graph.vineyard_id] = interactive_query
```
siyuan0322, line 943: If the interactive_instance creating eventually failed, should the item in dict be removed?

python/graphscope/framework/graph.py
```diff
@@ -156,6 +164,14 @@ def _close_learning_instances(self):
             instance.close()
         self._learning_instance_list.clear()
 
+    def _launch_interactive_instance_impl(self):
+        try:
+            sess = get_session_by_id(self.session_id)
+            sess.gremlin(self)
```
siyuan0322, line 170: Maybe we should add a option to control whether automatically create interactive instance whenever a graph is loaded
Reason:  If always create, when user want to get the interactive instance, they call `sess.gremlin(graph, params)`, and the method `sess.gremlin()` will simply find in the cache and return, the **params** is silently ignored, which may be confusing.  Also since user can never reach the logic of creating instance, it should be in another private method.

python/graphscope/client/session.py
```diff
@@ -913,20 +918,36 @@ def gremlin(self, graph, engine_params=None):
...
+            interactive_query = self._interactive_instance_dict[graph.vineyard_id]
+            if interactive_query.status == "running":
+                return interactive_query
+            elif interactive_query.status == "pending":
+                begin_time = time.time()
+                while True:
+                    time.sleep(1)
+                    if interactive_query.status == "running":
+                        return interactive_query
+                    if time.time() - begin_time >= timeout_seconds:
```
siyuan0322, line 930: After timeout, is the creating thread still alive? There can be chances that the starting procedure actually takes 70s, for example. What will happen if this method already raised timeout, and after sometime the thread successfully launched the server?

python/graphscope/client/session.py
```diff
@@ -894,14 +894,19 @@ def group_property_types(props):
...
 
+        Note that this method will be executed implicitly when a property graph created.
+
         Args:
             graph (:class:`Graph`): Use the graph to create interactive instance.
             engine_params (dict, optional): Configure startup parameters of interactive engine.
+                You can also configure this param by `graphscope.set_option(engine_params={})`.
                 See a list of configurable keys in
                 `interactive_engine/deploy/docker/dockerfile/executor.vineyard.properties`
+            timeout_seconds (int, optional): For waiting gremlin service ready. Defaults to 60.
```
siyuan0322, line 909: Since a gremlin is automatically called when graph is loaded, how can user configure the timeout parameter?

lidongze0629: graphscope.set_option(timeout_seconds=100)

python/graphscope/framework/graph.py
siyuan0322: There may be another circumstance that user can close a previous instance, and create a new one. We allow that. And this procedure of course will takes about ~30s. To prevent user wondering why two gremlin call consumed very different time, I think you should add the silent thread creating technique to somewhere in the doc, and in the session/graph comments.

python/graphscope/client/session.py
siyuan0322: If this is the case, maybe the actual timeout value is 600 by default. See in:
https://github.com/alibaba/GraphScope/blob/10f36b1dee462bf3efe9439a354e4f9e6d9f6da1/python/graphscope/config.py#L80

python/graphscope/client/session.py
lidongze0629: The `timeout` here is a little bit confused. `while True` is ok, as the status is either running or failed eventually after timeout.
```python
while True:
    time.sleep(1)
    if interactive_query.status == InteractiveQueryStatus.Running
        return interactive_query
    elif interactive_query.status == InteractiveQueryStatus.Failed:
        raise InteractiveEngineInternalError(interactive_query.error_msg)
```

python/graphscope/framework/graph.py
lidongze0629: Record error msg in `InteractiveQuery` when launching failed.

python/graphscope/interactive/query.py
lidongze0629: Done

python/graphscope/client/session.py
lidongze0629: Will set the status to `Failed` if create failed, So we rethrow exception when users call .gremlin() again.

python/graphscope/framework/graph.py
lidongze0629: Add `auto_create_interactive_engine`  in config.py. Defaults to True.

python/graphscope/config.py
```diff
@@ -80,4 +80,9 @@ class GSConfig(object):
     # GIE engine params
     engine_params = None
 
+    # GIE instance will be created automatically when a property graph loaded.
+    # Otherwise, you should create a GIE instance manually by `sess.gremlin` if
+    # `auto_create_interactive_engine` is False
+    auto_create_interactive_engine = True
```
sighingnow, line 86: naming: `initializing_interactive_engine` would be better.

But current option name is fine as well.

sighingnow (approved):
Now looks good enough to me. Thanks for the effort 👍 .

python/graphscope/interactive/__init__.py
```diff
@@ -18,8 +18,13 @@
...
```
sighingnow, line 27: Why restore this change?

python/graphscope/config.py
```diff
@@ -77,4 +77,12 @@ class GSConfig(object):
...
     log_level = "INFO"
 
+    # GIE engine params
+    engine_params = None
+
+    # GIE instance will be created automatically when a property graph loaded.
+    # Otherwise, you should create a GIE instance manually by `sess.gremlin` if
+    # `initializing_interactive_engine` is False
+    initializing_interactive_engine = True
+
```
sighingnow, line 87: I'm wondering if the `GSConfig` docs could be rendered in sphinx properly.

python/graphscope/framework/graph.py
siyuan0322: >  To prevent user wondering why two gremlin call consumed very different time, I think you should add the silent thread creating technique to somewhere in the doc, and in the session/graph comments.

Apart from the doc, when user run `.gremlin(sess, params)` with params, (we know the params are ignored), maybe we should print a warning to remind user that the params is not taken effect. And let them know the backend behavior.

python/graphscope/interactive/query.py
```diff
@@ -69,9 +85,29 @@ def graph_url(self):
...
+        self._error_msg = error_msg
+
+    def set_frontend(self, front_ip, front_port):
+        self._graph_url = "ws://%s:%d/gremlin" % (front_ip, front_port)
+        self._client = Client(self._graph_url, "g")
+
     def closed(self):
         """Return if the current instance is closed."""
-        return self._closed
+        return self._status == InteractiveQueryStatus.Closed
```
siyuan0322, line 110: Since you have three status type, what will happen if the status is `initializing` and `execute` is called?
Will the call be blocked? or error out?
ref: https://github.com/alibaba/GraphScope/blob/10f36b1dee462bf3efe9439a354e4f9e6d9f6da1/python/graphscope/interactive/query.py#L139-L141

python/graphscope/config.py
lidongze0629: We didn't expose `GSConfig` in Doc. Instead, `graphscope.set_option` and `graphscope.get_option` exposed.

https://graphscope.io/docs/reference/generated/graphscope.set_option.html#graphscope.set_option

python/graphscope/interactive/query.py
lidongze0629: `excute` `subgraph` and `traversal_source` can only run under `running` status. Otherwise, raise an exception.

python/graphscope/framework/graph.py
lidongze0629: Nice advice. I have updated doc of `gremlin`, which clarify how to create a new instance under the same graph with different engine_params. 
I didn't add `logger.warning` in code, as `gremlin` may run in a thread and hidden from users.

siyuan0322 (approved):

python/graphscope/framework/graph.py
lidongze0629: A new issue opened to trace InteractiveQuery create failed after close.

PR #175, opened 2021-03-03
fix issue of union result parser

shirly121:
## What do these changes do?

Fix for the new compiler of GIE, which combines the results returned by multiple streams.

## Related issue number

Fixes


lidongze0629 (approved):
Fix for the new compiler of GIE, which combines the results returned by multiple streams.

PR #179, opened 2021-03-08
Add edges to existing graph.

siyuan0322:
Add edges to existing graph.

Must merge alibaba/libvineyard#222 first and make a image.


sighingnow (changes requested):
Update the requirements.txt as well.

sighingnow (approved):

PR #192, opened 2021-03-15
Add preemptive param in session to give pod a good chance of being sc…

lidongze0629:
…heduled.

## What do these changes do?

This PR adds a new param named `preemptive` in session to give the pod a good chance of being scheduled.
 preemptive (bool, optional):
 If True, GraphScope instance will treat resource params (e.g. k8s_coordinator_cpu) as limits and provide the minimum available value as requests, but this will make pod has a `Burstable` QOS, which can be preempted by other pods with high QOS. Otherwise, it will set both requests and limits with the same value.

### DIFF
--- a/python/graphscope/deploy/kubernetes/resource_builder.py
+++ b/python/graphscope/deploy/kubernetes/resource_builder.py
@@ -697,7 +697,7 @@ class GSEtcdBuilder(DeploymentBuilder):
     """Builder for graphscope etcd."""
 
     _requests_cpu = 1.0
-    _requests_mem = "512Mi"
+    _requests_mem = "256Mi"
 
     def __init__(self, name, labels, image_pull_policy, replicas=1):
         self._name = name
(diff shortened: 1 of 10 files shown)


sighingnow (approved):
LGTM.

siyuan0322 (approved):

python/graphscope/deploy/kubernetes/resource_builder.py
```diff
@@ -673,6 +696,9 @@ def add_engine_container(self, name, image, cpu, mem, **kwargs):
 class GSEtcdBuilder(DeploymentBuilder):
     """Builder for graphscope etcd."""
 
+    _requests_cpu = 1.0
```
siyuan0322, line 699: https://github.com/alibaba/GraphScope/blob/f807eef2a37aae7724670c688feb05a8302c9c29/python/graphscope/config.py#L50
So in default case, the etcd limits cpu will be **0.5**, which is lower than the default and nonchangeable **1.0**.
I'm not sure, seems that means there will be error if using all default parameter to create the session?

PR #211, opened 2021-03-31
Add a general project operator.

### LINKED ISSUE #134: Project fragment: select a set of labels and properties to project

We have `Graph.project_to_simple` which select an edge + a vertice and a property to project a property fragment as a simple fragment. But we cannot obtain a subset of the original graph as a new property graph. The implementation should be straightforward.

Implements `Graph.project(elabels: Map[string, List[string]], vlabels: Map[string, List[string]]) -> Graph`.

- [x] Use `session.g()` to return a `Graph`.
- [x] Delete `remove` related syntax and docs.
- [ ] Unify `project` and `project_to_simple`
- [ ] Make `add_column` works by checking individual vertex label's signature, not by graph object id.

siyuan0322:
## What do these changes do?

1. Add a method `project` which can project a set of vertex labels and edge labels to a new property graph.
2. Make `project_to_simple` implicitly be called when an app that can only run over a simple graph is using that property graph.
3. `add_column` now performing validity check on vertex label level.

## Related issue number

Fixes #134


docs/graph_transformation.rst
```diff
@@ -0,0 +1,149 @@
...
+
+.. code:: python
+
+    import graphscope
+    from pathlib import Path
+    from graphscope.framework.loader import Loader
+
+    sess = graphscope.session()
+
+    graph = sess.g(directed=directed)
```
acezen, line 49: should I(the user) assign a `directed` variable before the statement ? because I don't know where the `directed` from in the example

siyuan0322: The `directed` here is not necessary, I copied it from the `ldbc.py` :). 
But the `sess.g()` and parameters is covered in `loading_graphs.rst`.

sighingnow (approved):

PR #233, opened 2021-04-06
Add template for frontend

siyuan0322:


sighingnow:
Why such a rename?

siyuan0322:
> Why such a rename?

Enable user to load another graph.

Issue #206, opened 2021-03-30
Add test cases for the optimized API for graph loading/projection


yecol:
docs for loading graph: https://graphscope.io/docs/loading_graph.html
proposal for graph projection: https://github.com/alibaba/GraphScope/issues/134

sighingnow:
Closed via #215 and #234.

PR #234, opened 2021-04-06
Fix bugs of project/add column listed by #215

siyuan0322:
Fix several bugs:
1. Dump schema when project.
2. Check existence before `get_edge_relationship`.
3. Load immediately after `add_column`.

related issue: #206 
related pr: #215

### DIFF
--- a/python/graphscope/framework/graph.py
+++ b/python/graphscope/framework/graph.py
@@ -432,6 +432,7 @@ class Graph(object):
         op = dag_utils.add_column(self, results, selector)
         graph = Graph(self._session, op)
         graph._base_graph = self
+        graph._ensure_loaded()
         return graph
 
     def to_numpy(self, selector, vertex_range=None):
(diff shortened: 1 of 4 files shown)


python/graphscope/framework/graph.py
```diff
@@ -432,6 +432,7 @@ def add_column(self, results, selector):
         op = dag_utils.add_column(self, results, selector)
         graph = Graph(self._session, op)
         graph._base_graph = self
+        graph._ensure_loaded()
```
sighingnow, line 435: Why `add_column` requires such force loading however `project` (as well as `project_to_simple`) doesn't need such hacking?

siyuan0322: It's @acezen thinks that the error should be raised in the method, not later. and `_project_to_simple` does have some checking as it has more information in the python side.

sighingnow: I see.

Leave a note/comment here would be better.

acezen:
LGTM

acezen (approved):

Issue #221, opened 2021-04-02
Remaining issues for standalone deployment


lidongze0629:
**TODO List**
- [x] ~~Replace zookeeper with [zetcd](https://github.com/etcd-io/zetcd) (waiting for refactor of GIE)~~
- [ ] Support install graphscope with ubuntu `apt` package manager. `sudo apt install graphscope | apt install graphscope-dev`
- [x] Support launch multiple session/instance (OnGoing dongze)
- [x] Add runtime directory to store runtime info (properties/logs)
- [ ] Multiple executors could get fragment from the same vineyard ? (need to discuess)
- [x] Release docker image and rich the documentation
- [ ] ~~Migrate existing GAE test from docker to physical machine (ubuntu20)~~
- [x] Add a cmake target to build and install GraphScope(GAE GIE GLE) into specified directory
- [x] ~~Delete GraphManager Role (integrate into coordinator)~~

I am sure it is far from perfect... and welcome any suggestions and features.

lidongze0629:
**Some thoughts of GIE logs**

1. Support streaming coordinator/frontend/executor logs to client. (Now logs(stdout/stderr) were redirected to disk file)
2. Support configure log path in running time. (Now determined at compile time)

lidongze0629:
**Docker images**
```shell
sudo docker pull registry.cn-hongkong.aliyuncs.com/graphscope/graphscope-runtime:ubuntu
# with vineyard 0.0.14
sudo docker pull registry.cn-hongkong.aliyuncs.com/graphscope/graphscope-vineyard:ubuntu-0.14
```

lidongze0629:
New issues to track these two remaining problems:

#243 
#701

PR #240, opened 2021-04-08
Add makefile target to simplify process of compilation

lidongze0629:
## Related issue number

Related to #221


Makefile
```diff
@@ -26,6 +33,58 @@ interactive_manager:
 push:
 	$(MAKE) -C $(WORKING_DIR)/k8s/ push
 
+.PHONY: install
+install: gle client coordinator gae gie
+
+.PHONY: client
+client:
```
sighingnow, line 40: The target `client` should depends on `gle`.

sighingnow (approved):

Makefile
lidongze0629: Modified
```shell
.PHONY: client
client: gle

.PHONY: coordinator
coordinator: client
```

PR #304, opened 2021-05-13
[GAIA] fix the bug of early stop optimization

### LINKED ISSUE #281: [BUG] Bug of early stop optimization in GAIA
**Describe the bug**
Bug query: 
```
g.V($id).repeat(both('KNOWS')).times(5).dedup().limit(10)
```
Queries like this should be fast completed due to early stop optimization. 
But such queries could be timeout sometimes.
There are bugs in the early stop mechanism in GAIA.

**To Reproduce**
Run the query above, and the bugs can only be reproduced accidentally.

youyangy:
## What do these changes do?
Fix the bug of early stop optimization.

## Related issue number

Fixes #281


sighingnow:
This part of code is beyond my knowedge thus I'm ccing correct reviewers...

sighingnow (approved):
And thanks for fixing the copyright header, that part looks good to me 👍

longbinlai (changes requested):
Please add corresponding test cases.

longbinlai (approved):

Issue #326, opened 2021-05-18
Benchmarking GAIA using typical Gremlin queries


youyangy:
1. Update benchmark tool to support typical gremlin queries: queries with sub-traversal, queries can be early-stopped and k-hop queries.
2. Benchmark report of above queries using current version GAIA as the baseline.
3. Fix inconsistent between parameter and query template files.

PR #339, opened 2021-05-24
[GAIA] update benchmark for typical queries and inconsistencies fix

youyangy:
## What do these changes do?
update benchmark for typical queries and inconsistencies fix

## Related issue number
part of #326


longbinlai (approved):
This has fixed consistency issues of ldbc benchmark.

PR #369, opened 2021-06-03
[GAIA] add encode for single event

### LINKED ISSUE #368: [BUG] Encode for single event is unsupported
**Describe the bug**
Encode trait for a single event is unsupported, and is used in distributed execution.

**To Reproduce**
Gremlin query with `limit` operator triggering early-stop can trigger this bug in distributed execution.

youyangy:
## What do these changes do?

add encode for single event

## Related issue number
Fixes #368

### DIFF
--- a/research/gaia/pegasus/pegasus/src/event/io/mod.rs
+++ b/research/gaia/pegasus/pegasus/src/event/io/mod.rs
@@ -92,7 +92,10 @@ impl Encode for Events {
     fn write_to<W: WriteExt>(&self, writer: &mut W) -> ::std::io::Result<()> {
         match self {
             &Events::Batched(ref events) => events.write_to(writer),
-            _ => unimplemented!("[Unreachable]"),
+            &Events::Single(ref event) => {
+                writer.write_u32(1u32)?;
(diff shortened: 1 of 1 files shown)


research/gaia/pegasus/pegasus/src/event/io/mod.rs
```diff
@@ -92,7 +92,10 @@ impl Encode for Events {
     fn write_to<W: WriteExt>(&self, writer: &mut W) -> ::std::io::Result<()> {
         match self {
             &Events::Batched(ref events) => events.write_to(writer),
-            _ => unimplemented!("[Unreachable]"),
+            &Events::Single(ref event) => {
```
longbinlai, line 95: You do not need to write `writer.write_u32()` for the Batched case? Otherwise how can you handle read. In addition, it looks like write_u8 is enough.

longbinlai (commented):
Can you describe more for this bug fix? How this happen, and why the codes can fix the bug.

longbinlai (approved):
This has fixed the single-event issue related to early-stop signal.

PR #372, opened 2021-06-04
Revert "Add closeness_centrality app and test case of duplicated mode…

acezen:


acezen:
close since the ci is fix.

Issue #303, opened 2021-05-13
Lazy and eager mode support


yecol:
Currently GraphScope only support run eager mode. 
We'd like to introduce a parameter `mode` in the session to support lazy execution.

Major changes: 
- functions in a lazy session, return a DAG node, rather than a real object;
- a DAG belongs to a session
- When in lazy mode, only results assigned in the `run` and their dependencies are reserved, other data is destroyed.
- defer the lazy support for gremlin.
- lazy session doesn't finalize when run, since results maybe in vineyard.

```python

# Explicitly open a session, the mode is eager by default.
s1 = graphscope.session(RESOURCES_GROUP_1)
s2 = graphscope.session(RESOURCES_GROUP_2)

# Build a graph
g1 = s1.g() # g1 belongs to s1,  return a DAG node
g2 = g1.add_vertices() # g2 is a DAG node and it inherits g1, session = s1
g3 = g2.add_edges() # g3.session = s1

# Create an interactive query
interactive = s2.gremlin(g3) # ERROR! g3 and s2 are not in the same session;
interactive = graphscope.gremlin(g3) # OK, inherits the session, s1.
ret = graphscope.bfs(g3, src=0) # s1.
lg = graphscope.learning(g3, label...) # s1

# lazy-mode
s3 = graphscope.session(mode="lazy")
# blabla... natually, these statements are not executed until triggers`run`
s3.run()

# ithis should be error:
#--------VERY-BEGINNING-OF-THE-SCRIPT
g = graphscope.g() # error, since there is no session, gs doesn't know where to evaluate this statement.

# session can be explicitly claimed as default. as before.

#--------VERY-BEGINNING-OF-THE-SCRIPT
s4 = graphscope.session(mode="lazy").as_default()
g4 = graphscope.g() 
g5 = g4.add_vertices()
g6 = g5.add_edges()

ret = graphscope.bfs(g6, src=0) 
r = s4.run(ret) # r can be used by others. the session is not finalized, since r maybe in v6d.




#--------VERY-BEGINNING-OF-THE-SCRIPT
s5 = graphscope.session(mode="lazy").as_default()
g1 = graphscope.g() # g1 is DagNode
g2 = sess.run(g1) # g2 is also a DagNode，but has been run. 

r1 = pagerank(g1) 
sess.run(r1) # reload g1

r2 = pagerank(g2)
sess.run(r2)  # don't need to relead g2

# --------NETWORKX-AND-SESSION

[...]

PR #377, opened 2021-06-07
[networkx] Adapt the networkx import to eager/lazy session mode 

acezen:
## What do these changes do?
Revise the networkx module import base on discussion in #303 

support two ways to import networkx module:
```python
# binding default_session to nx
import graphscope as nx 

# import from session
nx = sess.nx()
```

if related session is not eager mode, raise Error.

## Related issue number
#303 

Fixes

### DIFF
--- a/python/graphscope/nx/__init__.py
+++ b/python/graphscope/nx/__init__.py
@@ -34,6 +34,11 @@ from graphscope.nx.utils import *
 
 try:
     session = get_default_session()
+    if not session.eager():
+        raise ValueError(
+            "Networkx module need session to be eager mode. "
+            "The default session is lazy mode."
+        )
 except RuntimeError:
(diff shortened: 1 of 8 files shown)


python/graphscope/nx/__init__.py
```diff
@@ -29,3 +31,15 @@
...
+    session = get_default_session()
+    if not session.eager():
+        raise ValueError(
+            "Networkx module need session to be eager mode. "
+            "The default session is lazy mode."
+        )
+except RuntimeError:
+    # no default session found.
+    session = None
+setattr(Graph, "_session", session)
```
sighingnow, line 45: This part is not used anymore.

sighingnow (approved):
LGTM.

PR #401, opened 2021-06-10
Ensure graphscope.proto preponderate over outside `proto` directory.

### LINKED ISSUE #400: [BUG] `import graphscope` may when there's a `proto` directory in the root `path`.

sighingnow:
As some other packages (i.e., `proto-plus-python`) may install a `proto` directory in the root path.

## Related issue number

Fixes #400.


siyuan0322 (approved):
👍

PR #403, opened 2021-06-10
Check version compatibility during session connecting

lidongze0629:
## Related issue number

Fixes

### DIFF
--- a/coordinator/gscoordinator/coordinator.py
+++ b/coordinator/gscoordinator/coordinator.py
@@ -38,6 +38,7 @@ from concurrent import futures
 from io import StringIO
 
 import grpc
+from packaging import version
 
 from gscoordinator.io_utils import StdoutWrapper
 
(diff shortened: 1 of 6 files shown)


coordinator/gscoordinator/coordinator.py
```diff
@@ -214,6 +214,14 @@ def ConnectSession(self, request, context):
         self._streaming_logs = True
         sys.stdout.drop(False)
 
+        # check version compatibility from client
+        if self._request.version != __version__:
+            logger.warning(
+                "Version between client and server is inconsistent: %s vs %s",
+                self._request.version,
+                __version__,
+            )
```
sighingnow, line 223: We should follow the semver, and when only the patch version differ it SHOULD still could work.

python/graphscope/client/session.py
```diff
@@ -735,6 +735,8 @@ def close(self):
                     instance.close()
             except InteractiveEngineInternalError:
                 pass
+            except Exception:
+                pass
```
sighingnow, line 739: Just catch `Exception` should be enough, as it is a baseclass of `InteractiveEngineInternalError `.

python/graphscope/client/session.py
```diff
@@ -744,10 +746,15 @@ def close(self):
                     instance.close()
             except LearningEngineInternalError:
                 pass
+            except Exception:
+                pass
         self._learning_instance_dict.clear()
```
sighingnow, line 751: Ditto.

sighingnow (changes requested):

sighingnow (approved):

Issue #415, opened 2021-06-16
Fix gremlin-test on persistent storage


tianliplus:
as titled

PR #416, opened 2021-06-16
fix gremlin test case

tianliplus:
## What do these changes do?

fix bug in get_partition_id() of local store

## Related issue number

#415


siyuan0322 (approved):

tianliplus:
fix bug in get_partition_id() of local store

PR #422, opened 2021-06-21
Integrate GAIA with GraphScope Storage (Read Path)

### LINKED ISSUE #417: Integrate GAIA with GraphScope Storage (Read Path)
Integrate GAIA with GraphScope Storage on read path

Adapt store access in GAIA to GlobalGraphQuery in GraphScope

BingqingLyu:
## Related issue number

Fixes #417


longbinlai (commented):
Way too many code duplicates. Please consider abstracting common-used functions and also import necessary tools instead of directly copying the codes.

sighingnow:
Could we put this part to `research/gaia/gs` rather than `/research/gaia_gs`, as it is not a parallel component with gaia.

interactive_engine/src/executor/store/Cargo.toml
```diff
@@ -26,6 +26,7 @@ maxgraph-common = { path = "../../common/rust/common" }
 
 [dependencies.rocksdb]
 git = "https://github.com/pingcap/rust-rocksdb.git"
+rev = "23a604459899f4f7a4c8967f16b6fa623474f27b"
```
longbinlai, line 29: Why relies on some `rev` here?

research/gaia/gremlin/gs_gremlin/src/bin/start_rpc_server.rs
```diff
@@ -0,0 +1,108 @@
...
+    let num_servers = if let Some(h) = &host_config {
+        h.peers.len()
+    } else {
+        1
+    };
+    let config = combine_config(server_config.server_id, host_config, common_config);
+    let addr = format!("{}:{}", "0.0.0.0", server_config.rpc_port);
+
+    let graph = build_modern_mock_graph();
+    let partitioner = graph.clone();
```
longbinlai, line 91: Not sure what this means, `partitioner = graph.clone()`? Why a partitioner is cloning a graph?

interactive_engine/src/executor/store/Cargo.toml
BingqingLyu: This is because the new commits of rust-rocksdb have introduced an incompatible function we used, which leads to compile failure. So I add the  commit id from the Cargo.lock for compatible (with older version) consideration.

BingqingLyu: This is because the new commits of rust-rocksdb have introduced an incompatible function we used, which leads to compile failure. So I add the  commit id from the Cargo.lock for compatible (with older version) consideration.

research/gaia/gremlin/gs_gremlin/src/bin/start_rpc_server.rs
BingqingLyu: This is just an example (and will be replaced when integrate with graphscope storage). I have implemented GraphPartitionManager trait (from graphscope) for MockGraph, following the way  graphscope storage does it.

interactive_engine/src/executor/store/Cargo.toml
longbinlai: This is very odd. Instead, fork a github version to GraphScope to stay in whatever version we'd like is a more common practice.

research/gaia/gremlin/gs_gremlin/src/bin/start_rpc_server.rs
longbinlai: If this is the case, a recommendation is to mark this pull request as [WIP], and use this temporary MockGraph in your own test instead of submitting to the code bases. I've tried to review the codes, but MockGraph-related codes introduce too many distractions (and to be honest, I find it hard to understand some codes).

BingqingLyu: Ok, I have removed the start_rpc_server and the mock_graph related files, which are for test.

research/gaia/gremlin/gremlin_core/src/graph_proxy/storage.rs
```diff
@@ -336,30 +292,24 @@ impl GraphProxy for DemoGraph {
...
     let details = LazyVertexDetails::new(v.get_id(), store);
-    let id = encode_runtime_v_id(&v);
-    let label = encode_runtime_v_label(&v);
+    let id = v.get_id() as ID;
+    let label = Some(Label::Id(v.get_label()[0]));
     Vertex::new(id, label, details)
 }
 
+#[inline]
 fn to_runtime_vertex_with_property(v: LocalVertex<DefaultId>, props: &Vec<String>) -> Vertex {
```
longbinlai, line 307: How can you represent no copying any property? Better take a `Option<&Vec<String>>`

PR #431, opened 2021-06-23
[graphscope] add mapping of partition id to server id in GraphPartitionManager

BingqingLyu:
Add a function in GraphPartitionManager for GAIA Integration: 
`fn get_server_id(&self, pid: PartitionId) -> i32;`

### DIFF
--- a/interactive_engine/src/executor/runtime/src/store/ffi.rs
+++ b/interactive_engine/src/executor/runtime/src/store/ffi.rs
@@ -1411,6 +1411,10 @@ impl GraphPartitionManager for VineyardPartitionManager {
         }
     }
 
+    fn get_server_id(&self, _pid: u32) -> i32 {
+        unimplemented!()
+    }
+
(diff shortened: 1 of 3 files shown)


tianliplus (approved):

interactive_engine/src/executor/runtime/src/store/ffi.rs
```diff
@@ -1411,6 +1411,10 @@ impl GraphPartitionManager for VineyardPartitionManager {
         }
     }
 
+    fn get_server_id(&self, _pid: u32) -> i32 {
+        unimplemented!()
```
longbinlai, line 1415: unimplemented!()? so must follow with `TODO(who_owns_this) ` . If this code will never be used and implemented, should be `unreachable!()`

BingqingLyu: done

PR #451, opened 2021-06-29
Support `with` keyword in session

### LINKED ISSUE #393: Support use `with` keyword with session to enable automatically release resources.
Support context manage for `Session`.
```python

with graphscope.session() as sess:
    pass
# automatically close `sess` and release sources.

```

siyuan0322:
## What do these changes do?
This PR enables `with` keyword in session, which can automatically release resourced when the context manager exits.
Moreover, use `with` will automatically set the session as the default session.

Two simple test cases has been added that can fully demonstrate its usage.

## Related issue number

Fixes #393

### DIFF
--- a/python/graphscope/client/session.py
+++ b/python/graphscope/client/session.py
@@ -695,6 +695,10 @@ class Session(object):
         info["engine_config"] = self._engine_config
         return info
 
+    @property
+    def closed(self):
+        return self._closed
+
     def eager(self):
         return self._config_params["mode"] == "eager"
 
@@ -781,14 +785,23 @@ class Session(object):
         except Exception:  # pylint: disable=broad-except
             pass
 
+    def _check_closed(self, msg=None):
+        """Internal: raise a ValueError if session is closed"""
+        if self.closed:
+            raise ValueError("Operation on closed session." if msg is None else msg)
+
+    # Context manager
     def __enter__(self):
-        """Register self as default session"""
+        """Context management protocol.
+        Returns self and register self as default session.
+        """
+        self._check_closed()
         self.as_default()
         return self
 
     def __exit__(self, type, value, traceback):
         """Deregister self from the default session,
-        close the session and release the resources.
+        close the session and release the resources, ignore all exceptions in close().
         """
         try:
             self._deregister_default()
--- a/python/graphscope/client/session.py
+++ b/python/graphscope/client/session.py
@@ -781,6 +781,21 @@ class Session(object):
         except Exception:  # pylint: disable=broad-except
             pass
 
+    def __enter__(self):
+        """Register self as default session"""
+        self.as_default()
+        return self
+
+    def __exit__(self, type, value, traceback):
+        """Deregister self from the default session,
+        close the session and release the resources.
+        """
+        try:
+            self._deregister_default()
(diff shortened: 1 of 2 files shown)


python/tests/test_session.py
```diff
@@ -255,3 +255,13 @@ def test_border_cases():
...
+
+
+def test_with():
+    with graphscope.session(cluster_type="hosts") as sess:
+        assert graphscope.get_default_session() == sess
+
+    sess = graphscope.session(cluster_type="hosts")
+    with sess:
+        pass
+    assert sess.info["status"] == "closed"
```
sighingnow, line 267: This test case, as well as the `self.close()` in `__exit__` looks strange.

What's the behavior of `with <established session>` in other frameworks, e.g., tensorflow?

python/graphscope/client/session.py
```diff
@@ -781,6 +781,21 @@ def __del__(self):
...
+
+    def __exit__(self, type, value, traceback):
+        """Deregister self from the default session,
+        close the session and release the resources.
+        """
+        try:
+            self._deregister_default()
+            self.close()
+        except Exception:
+            pass
```
sighingnow, line 797: No need to this catch, as we already use `except Exception` to catch all in `self.close()` itself.

python/tests/test_session.py
siyuan0322: It doesn't close session.

python/graphscope/client/session.py
siyuan0322: No, `self.close()` doesn't do `except Exception`, it's in `__del__`

python/tests/test_session.py
sighingnow: `sess.info["status"] == "closed"`?

python/graphscope/client/session.py
sighingnow: I see, then it looks good to me.

python/tests/test_session.py
siyuan0322: My understanding is that our primary goal using `with` is to automatically release resources, is that correct?

siyuan0322: Any problem with the `sess.info["status"] == "closed"`? Could you make it more clear 😂

sighingnow: With `with expression` statement the `__enter__` and `__exit__` should be symmetry, i.e., `open(....)` in `__enter__` and `close(...)` in `__exit__`. The `expression` it self shouldn't handle such stuffs.

If we `.as_default()` in `__enter__` then I think the expected behavior in `__exit__` should be _pop_ the default session, rather than closing it.

siyuan0322: But the open stage is done in `__init__`, the way you are saying is to solely do the register default and deregister stuff.

> If we .as_default() in __enter__ then I think the expected behavior in __exit__ should be pop the default session, rather than closing it.
Yes, I actually closed it after poped the default session.

siyuan0322: So the problem becomes, should we close the session in `__exit__`? What is already known is we don't open the session in `__enter__`.

sighingnow (approved):

PR #458, opened 2021-07-01
Add snapshot id when query store

BingqingLyu:
## What do these changes do?

Add snapshot id when query store

## Related issue number

Fixes


research/gaia/gremlin/gremlin_core/src/process/traversal/step/map/identity.rs
```diff
@@ -89,6 +89,7 @@ impl MapFuncGen for IdentityStep {
         let step = self.step;
         let mut params = QueryParams::new();
         params.set_props(step.required_properties);
+        params.snapshot_id = if let Some(si) = step.snapshot_id { Some(si.id) } else { None };
```
longbinlai, line 92:
```
params.snapshot_id = step.snapshot_id.map(|si| si.id )
```
Apply to all the following codes.

research/gaia/gremlin/proto/gremlin.proto
```diff
@@ -120,6 +124,7 @@ message GraphStep {
   repeated TraverserRequirement traverser_requirements = 5;
   // The properties of the entities that need to be saved
   PropKeys required_properties = 6;
+  SnapShotId snapshot_id = 7;
```
longbinlai, line 127: Add comments. Apply as follows.

research/gaia/gremlin/gremlin_core/src/process/traversal/step/map/identity.rs
BingqingLyu: Done

research/gaia/gremlin/proto/gremlin.proto
BingqingLyu: Done

research/gaia/gremlin/gremlin_core/src/structure/graph.rs
```diff
@@ -23,11 +23,12 @@ pub struct QueryParams<E: Element + Send + Sync> {
     pub limit: Option<usize>,
     pub props: Option<Vec<PropKey>>,
     pub filter: Option<Arc<Filter<E, ElementFilter>>>,
+    pub snapshot_id: Option<i64>,
```
longbinlai, line 26: It looks like MaxGraph already have a definition of `SnapshotId`., import it instead of directly using `i64`

PR #467, opened 2021-07-05
Fix string oid graph transformation bug

acezen:
## What do these changes do?

- no need to call the vertex map construct, since every worker has the complete global vertex map.


acezen:
@pwrliang I have reverted the parallel edge and vertex code, can't you take a look again?

lidongze0629 (approved):

PR #478, opened 2021-07-06
Set GRPC max message length options in coordinator server side

lidongze0629:
## Related issue number

Fixes


sighingnow (approved):
LGTM.

PR #507, opened 2021-07-09
Set decorator>=5.0.5 since it's old version(4.4.2) may not compatible with NetworkX latest release

acezen:
can reproduce with NetworkX 2.6.1 and decorator 4.4.2
```python
from networkx.utils.decorators import not_implemented_for
@not_implemented_for("directed")
 def sp_function(G):
      pass

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-5-b6e5bef0b8d8> in <module>
----> 1 @not_implemented_for("directed")
      2 def sp_function(G):
      3     pass
      4

~/.local/anaconda3/envs/graphscope/lib/python3.8/site-packages/networkx/utils/decorators.py in not_implemented_for(*graph_types)
     89         return g
     90
---> 91     return argmap(_not_implemented_for, 0)
     92
     93

TypeError: __init__() missing 1 required keyword-only argument: 'try_finally'
      
```

### DIFF
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -5,7 +5,7 @@ grpcio
 grpcio-tools
 kubernetes
 nest_asyncio
-networkx
+networkx<2.6
 numpy
 pandas
 protobuf>=3.12.0
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -5,7 +5,7 @@ grpcio
 grpcio-tools
 kubernetes
 nest_asyncio
-networkx<2.6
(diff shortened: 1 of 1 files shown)


sighingnow:
Cannot reproduce with python 3.9.2 and network HEAD.

Which python are you working with?

acezen:
> Cannot reproduce with python 3.9.2 and network HEAD.
> 
> Which python are you working with?

Python 3.8

PR #530, opened 2021-07-14
Initialize job compiler in adapter for starting engine service

BingqingLyu:
## What do these changes do?

As titled


tianliplus (approved):
LGTM

PR #554, opened 2021-07-15
Fix macOS ci: change to llvm 8

acezen:
## What do these changes do?
In MacOS runner image version 20210712.4, the llvm@9 we use in ci thinks 'preadv' and 'pwritev' is enable to use but it's available until MacOS 11.0, so we downgrade  llvm to 8.


lidongze0629 (approved):
Nice Job!

PR #580, opened 2021-07-22
Coordinator get source to source json successful from gae compiler

lidongze0629:
## Related issue number

Fixes

### DIFF
--- a/coordinator/gscoordinator/object_manager.py
+++ b/coordinator/gscoordinator/object_manager.py
@@ -17,6 +17,25 @@
 #
 
 from gremlin_python.driver.client import Client
+from gremlin_python.driver.serializer import GraphSONMessageSerializer
+
+"""Patch for gremlin_python serializer to support "gae" processor
+"""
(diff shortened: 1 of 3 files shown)


coordinator/gscoordinator/object_manager.py
```diff
@@ -47,6 +66,19 @@ def __init__(self, key, frontend_endpoint, object_id):
         self.closed = False
 
     def submit(self, message, bindings=None, request_options=None):
+        if request_options is not None and "engine" in request_options:
+            from gremlin_python.driver import request
+
+            rm = request.RequestMessage(
+                processor="gae",
```
siyuan0322, line 73: As you would pass `request_options` like `{'engine': 'gae'}`, I think use `processor=request_options['engine']` would be better.

yecol: to make it runnable ASAP, it is acceptable, you may add a "FIXME".

PR #590, opened 2021-07-23
add vineyard adaptor

shirly121:
## Related issue number

Fixes


tianliplus (approved):
Adaptor should not depend on v2.  

Pass for now. Let's open another issue to fix the dependency problem.

siyuan0322 (approved):

PR #628, opened 2021-07-30
[Bug Fix]: add u128 in Primitive in Object

BingqingLyu:
## What do these changes do?

Fix a bug in where step, e.g.,  g.V(-1).as('a').out().in().where(eq('a')). 

The reason is that in where step, it will set the expected value (i.e., global id referred by tag 'a') as a type of Object (in this case it is Object::Blob). However, when it comes to comparing with the output of in() step, we have the actual global_id of type u128, and the type inference in rust will also regard the expected value as u128. Thus, the bug occurs when we try to get Object::Blob value as a u128 value.

The solution is to add u128 as a primitive in Object, since it is commonly used in runtime (as it is the type of global id).

## Related issue number

Fixes

### DIFF
--- a/research/gaia/dyn_type/src/object.rs
+++ b/research/gaia/dyn_type/src/object.rs
@@ -28,6 +28,7 @@ pub enum RawType {
     Byte,
     Integer,
     Long,
+    LLLong,
     Float,
     String,
     Blob(usize),
@@ -39,6 +40,7 @@ pub enum Primitives {
     Byte(i8),
     Integer(i32),
     Long(i64),
+    LLLong(u128),
     Float(f64),
 }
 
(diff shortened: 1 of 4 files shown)


research/gaia/dyn_type/src/object.rs
```diff
@@ -39,6 +40,7 @@ pub enum Primitives {
     Byte(i8),
     Integer(i32),
     Long(i64),
+    LLLong(u128),
```
longbinlai, line 43: Shall be `LLong(i128)`

BingqingLyu: Modified as ULLong(u128) for now

longbinlai (approved):

PR #668, opened 2021-08-09
support java16 && upgrade rust toolchain

### LINKED ISSUE #565: Compatibility for latest version of toolchain, like Java, cargo...
Currently GIE only supports Java 8.
We need to improve the compatibility for Java 8+.

- [x] Java: support latest? 
- [x] Cargo: only support clang version 8-10, however, the latest version is 12;
- [x] sed: depends on a variant, make macOS need to install a package `gnused`;
- [x] update the script for installing dependencies (@acezen)

tianliplus:
## What do these changes do?

support java16 && upgrade rust toolchain

## Related issue number

Fixes #565


yecol (approved):
Thanks!

PR #671, opened 2021-08-09
fix sending unvalidated data to a web browser

### LINKED ISSUE #607: Cross-Site Scripting: Reflected
https://github.com/alibaba/GraphScope/blob/1f62da38de65c28c3b27aac378562d0d5a4ef13f/interactive_engine/src/instance-manager/src/main/java/com/alibaba/maxgraph/admin/controller/InstanceManagerController.java#L153-L171

Sending unvalidated data to a web browser can result in the browser executing malicious code.

In line 153， ‘graphName’ is regarded to as tainted data,it will be Affect line 170.

### LINKED ISSUE #608: Command Injection
https://github.com/alibaba/GraphScope/blob/1f62da38de65c28c3b27aac378562d0d5a4ef13f/interactive_engine/src/instance-manager/src/main/java/com/alibaba/maxgraph/admin/controller/InstanceManagerController.java#L371-L381

Executing commands from an untrusted source or in an untrusted environment can cause an application to execute malicious commands on behalf of an attacker.

In line 371, ‘graphName’ is regarded to as tainted data,it will be affect line 381.

shirly121:
## Related issue number

Fixes #607 
Fixes #608

### DIFF
--- a/interactive_engine/src/instance-manager/src/main/java/com/alibaba/maxgraph/admin/controller/InstanceManagerController.java
+++ b/interactive_engine/src/instance-manager/src/main/java/com/alibaba/maxgraph/admin/controller/InstanceManagerController.java
@@ -100,6 +100,11 @@ public class InstanceManagerController {
         String stdoutMessage = "";
 
         try {
+            if (!isValidGraphName(graphName)) {
+                createInstanceEntity.setErrorCode(-1);
+                createInstanceEntity.setErrorMessage("Invalid graph name");
+                return createInstanceEntity;
+            }
             List<String> createCommandList = new ArrayList<>();
 
             createCommandList.add(instanceProperties.getCreateScript());
@@ -195,9 +200,9 @@ public class InstanceManagerController {
         int frontendPort = 0;
 
         String schemaPath = "/tmp/" + graphName + ".json";
-        if (StringUtils.isEmpty(SecurityUtil.pathFilter(schemaPath))) {
+        if (!isValidGraphName(graphName) || StringUtils.isEmpty(SecurityUtil.pathFilter(schemaPath))) {
             createInstanceEntity.setErrorCode(-1);
-            createInstanceEntity.setErrorMessage("Invalid graph name=>" + graphName);
+            createInstanceEntity.setErrorMessage("Invalid graph name");
             return createInstanceEntity;
         }
         try {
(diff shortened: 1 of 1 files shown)


interactive_engine/src/instance-manager/src/main/java/com/alibaba/maxgraph/admin/controller/InstanceManagerController.java
```diff
@@ -100,6 +100,11 @@ public CreateInstanceEntity createLocalInstance(@RequestParam("graphName") Strin
         String stdoutMessage = "";
 
         try {
+            if (!isValidGraphName(graphName)) {
+                createInstanceEntity.setErrorCode(-1);
+                createInstanceEntity.setErrorMessage("Invalid graph name");
```
siyuan0322, line 105: Better to put the graphName in the error message as well.

interactive_engine/src/instance-manager/src/main/java/com/alibaba/maxgraph/admin/controller/InstanceManagerController.java
```diff
@@ -431,4 +442,9 @@ public CloseInstanceEntity closeInstance(@RequestParam("graphName") String graph
     public String queryFrontendEndpoint(@RequestParam("graphName") String graphName) {
         return FrontendMemoryStorage.getFrontendStorage().getFrontendEndpoint(graphName).getFrontEndpoint();
     }
+
+    public boolean isValidGraphName(String graphName) {
+        String validPattern = "^\\w{1,128}$";
```
siyuan0322, line 447: Please give a comment to simply explain the regex.

interactive_engine/src/instance-manager/src/main/java/com/alibaba/maxgraph/admin/controller/InstanceManagerController.java
```diff
@@ -195,9 +200,9 @@ public CreateInstanceEntity createInstance(@RequestParam("graphName") String gra
         int frontendPort = 0;
 
         String schemaPath = "/tmp/" + graphName + ".json";
-        if (StringUtils.isEmpty(SecurityUtil.pathFilter(schemaPath))) {
+        if (!isValidGraphName(graphName) || StringUtils.isEmpty(SecurityUtil.pathFilter(schemaPath))) {
             createInstanceEntity.setErrorCode(-1);
-            createInstanceEntity.setErrorMessage("Invalid graph name=>" + graphName);
+            createInstanceEntity.setErrorMessage("Invalid graph name");
```
siyuan0322, line 205: Apparently there would be two reasons, one is the `graphName` is not valid, one is the `schemaPath` is empty. we should be able to distinguish them in the error message. So maybe put these validations in separate branches would be more clear.

siyuan0322 (approved):

Issue #660, opened 2021-08-06
The interactive engine frontend log is randomly output to a file .out/.log


acezen:
**Describe the bug**

As https://github.com/alibaba/GraphScope/blob/fcca5f2a01a92a9f3ba930a7e1af54a851de3337/interactive_engine/bin/giectl#L351-L355 shows, during the  cerate gremlin instance on local, we need to get the frontend port from log file.
But sometimes the frontend log was write out to stderr and failed to get the port here and raise `FRONTEND_PORT match failed` error

**To Reproduce**

- In github ci runner, the frontend log is output to .out file.
- But in `hongkong.aliyuncs.com/graphscope/graphscope-vineyard:ubuntu0.2.5` container, the frontend log is output to .err file

**Expected behavior**

The frontend log should be always output to frontend.log, not frontend.out or frontend.err.

**Environment (please complete the following information):**
 - GraphScope version: [main]
 - OS: [Ubuntu 20.04]

**Additional context**

I found the random output behavior was introduce by [commit](https://github.com/alibaba/GraphScope/commit/19f7d0b53683317ecb03d680673a4087bc5d0056#).
After remove `gaia-adaptor` dependency from pom.xml,  output act as expected.

PR #693, opened 2021-08-13
An ad-hoc solution to find frontend port from log file

acezen:
## What do these changes do?

As title

## Related issue number

related #660


siyuan0322 (approved):
Better to add a fixme direct to the issue, indicating it's a temporal solution

PR #702, opened 2021-08-16
Remove dependency of v2 from gaia-adaptor

shirly121:
## Related issue number

Fixes


interactive_engine/src/tree
```diff
@@ -0,0 +1,2026 @@
+[INFO] Scanning for projects...
```
yecol, line 1: please do not add this log file to repo.

yecol (changes requested):

PR #713, opened 2021-08-19
Adapt to latest vineyard: the `Create()` API has changed.

sighingnow:
## What do these changes do?

The static `Create()` method returns a `std::unique_ptr` instead.

## Related issue number

N/A


siyuan0322 (approved):
👍

PR #717, opened 2021-08-23
Fix gremlin hang

### LINKED ISSUE #692: [BUG] build-gss job  sometimes get stuck in ci
**Describe the bug**
as title 

**Screenshots**
截屏2021-08-13 下午2 12 12

[related action](https://github.com/alibaba/GraphScope/runs/3309293080?check_suite_focus=true)

**Environment (please complete the following information):**
 - GraphScope version: [main]
 - OS: [Linux]

**Additional context**

tianliplus:
## Related issue number

Fixes #692

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,6 @@ on:
   push:
     branches:
       - main
-      - fix_gremlin_hang
     paths-ignore:
       - 'gnn_engine/**'
       - '**.md'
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -485,7 +485,7 @@ jobs:
             registry.cn-hongkong.aliyuncs.com/graphscope/maxgraph_standalone_manager:${{ github.sha }} || true
 
   build-gss:
-    runs-on: ubuntu-20.04
+    runs-on: self-hosted
```
siyuan0322, line 488: I think you should put a comment here, that the docker need to be configured doesn't need sudo priviledge on the host.  We may add new runners in the future.

acezen:
thanks for the fix. Can you put some comment about why use `self-hosted` can fix the issue.

tianliplus:
> thanks for the fix. Can you put some comment about why use `self-hosted` can fix the issue.

Because this error only happend on github runner. It is env-related and hard to debug. So migrate to `self-hosted` for now.

PR #743, opened 2021-08-30
Fix id encoding

### LINKED ISSUE #741: Gaia should use 64bit edge inner id on store_v2
image

This `id` should be 64 bit.

shirly121:
1. update edge id from string to bytes in pb
2. use edge inner id for v2 and vineyard as edge ID in gaia runtime
3. convert edge id into i64 from bytes for v2 in compiler

## Related issue number

Fixes #741


research/gaia/gremlin/gremlin_core/src/process/traversal/step/source.rs
```diff
@@ -169,15 +169,9 @@ pub fn graph_step_from(
...
-                        uid
-                    } else if let Ok(id) = i64::from_str(&id) {
-                        // in case that source id is negative
-                        id as ID
-                    } else {
-                        return Err("Parse source id failed")?;
-                    };
+                for id_bytes in opt.ids {
+                    let id = read_be_u128(&mut id_bytes.as_slice());
+                    debug!("source id_bytes: {:?}, id: {:?}", id_bytes, id);
```
siyuan0322, line 174: Will this line produce a lot of logs when the graph is huge?

siyuan0322 (commented):
Please remove the try except block here, only keep the line 133 so that we can see if the fix works.
https://github.com/alibaba/GraphScope/blob/9e8e066d213e31f32c9c77faedb631da113c410b/python/tests/kubernetes/test_store_service.py#L129-L136

research/gaia/gremlin/gremlin_core/src/process/traversal/step/source.rs
BingqingLyu: debug!() will only print the log when the log level is debug and I think the default log level is info? Besides, here it only prints the ids in query request, not the query results.

siyuan0322 (approved):
LGTM

interactive_engine/src/v2/src/main/java/com/alibaba/graphscope/gaia/MaxGraphStore.java
```diff
@@ -80,4 +82,24 @@ public long getSnapShotId() {
     public synchronized void updateSnapShotId() {
         this.cachedGraphSchemaPair.update();
     }
+
+    @Override
+    public Object getCompositeId(GremlinResult.Edge edge) {
```
tianliplus, line 87: Two edge can have same composite id?

PR #744, opened 2021-08-31
Fetch stdout and stderr log of grape engine separately

### LINKED ISSUE #729: Redirect stderr to client even when show_log=False

The grape-engine may complaint using `LOG(ERROR)` or `LOG(WARN)`, which are printed to stderr and handled by mpi. However when `show_log=False` when launching session, all logs are discarded.

The errors or warnings should be unconditionally printed, IMHO.

+ Fixes the `FLAGS_stderrthreshold` issue, by either
   - changing the threshold
   - add a custom sinker to `LOG(ERROR)`
+ Capture the logs on stderr, and redirect to client.

**Additional context**

acezen:
## What do these changes do?
 - This pull request revise the fetch log implement of client tp get stdout and stderr log separately. 
 - In engine side, we use a RedirectLogSink to process LOG(ERROR) and OTHERS LOG of glog separately.

## Related issue number
Fixes #729

### DIFF
--- a/analytical_engine/core/grape_engine.cc
+++ b/analytical_engine/core/grape_engine.cc
@@ -14,6 +14,7 @@
  */
 
 #include <csignal>
+#include <limits>
 #include <thread>
 #include <utility>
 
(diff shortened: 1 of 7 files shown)


analytical_engine/core/grape_engine.cc
```diff
@@ -146,9 +147,29 @@ extern "C" void master_signal_handler(int sig, siginfo_t* info, void* context) {
...
 }
 
+// A custom sinker to redirect glog messages to stdout/stderr.
+class RedirectLogSink : public google::LogSink {
+  virtual void send(google::LogSeverity severity, const char* full_filename,
+                    const char* base_filename, int line,
+                    const struct ::tm* tm_time, const char* message,
+                    size_t message_len) {
+    // we redirect GLOG_ERROR message to stderr, others to stdout
+    if (severity != google::GLOG_ERROR) {
```
sighingnow, line 157: `GLOG_WARNING` as well.

sighingnow: `GLOG_WARNING` should be routed to `std::cerr` as well.

analytical_engine/core/grape_engine.cc
```diff
@@ -146,9 +147,29 @@ extern "C" void master_signal_handler(int sig, siginfo_t* info, void* context) {
   }
 }
 
+// A custom sinker to redirect glog messages to stdout/stderr.
+class RedirectLogSink : public google::LogSink {
```
yecol, line 151: Are you sure we need a `redirectlogsink`? I think glog can be configured to send error/out to stdout/stderr?

acezen: glog can only set to send log to stderr, others would write to log file.
FYI: https://github.com/google/glog#setting-flags

yecol (approved):

PR #749, opened 2021-09-01
Output the graphscope-store log to both console  and file

### LINKED ISSUE #404: Improve the error handling of GraphScope-Store to make it more cloud-native.
The store works well when all the requirements are fulfilled and configured properly, but when error occurs the store will more likely to **stuck** or **swallow the exception** to a log file that only can be accessed in the pod, and no restart will be scheduled.

We should
- Fail the graphscope-store cluster when the maxgraph components are not starting up normally,
- Provide more error information to external user such as the stdout of pod so that kubernetes can gather those logs
- Provide tips for users about the potential ways to debug the error.

Moreover, some restart policy may be applied in case of occasional failures.

### LINKED ISSUE #518: Reorganize the scripts for GIE launching
It seems in the current version, the scripts for GIE are not well organized. 
I suggest we reorg the files in `/interactive_engine/distribution` and `/interactive_engine/deploy`

Currently, the file tree of distribution:
```
distribution
├── pom.xml 
└── src
    ├── assembly
    │   └── assembly.xml
    ├── bin 
    │   ├── admin_tools.sh
    │   ├── coordinator.sh
    │   ├── frontend.sh
    │   ├── ingestor.sh
    │   ├── load_tool.sh
    │   ├── max_node.sh
    │   ├── maxgraph-env
    │   ├── maxgraph.sh
    │   ├── startall.sh
    │   └── store.sh
    └── config
        ├── config.template
        ├── coordinator.config
        ├── frontend0.config
        ├── ingestor0.config
        ├── ingestor1.config
        ├── log4rs.yml
        ├── logback.xml
        ├── logback_debug.xml
        ├── sample.config
        ├── store0.config
        └── store1.config
```

problems:
- this folder only contains binary exetuables and configurations, thus it doesn't need to follow java code structure with a pom?
- the scripts in `bin` folder are quite simple, all of them are one-line command of Java. I think we can aggregate them to a single binary executable (e.g., `giectl`) with sub-command. like `giectl start-coordinator`, `giectl start-frontend`, `giectl start-all`...
- for configuration files, we should provide config template, rather than explicit ingestor0, ingestor1 frontend0 files.

For `interactive_engine/deploy` folder, currently the files are organized like this:
```
deploy
├── docker
│   └── dockerfile
│       ├── coordinator-entrypoint.sh
│       ├── coordinator.application.properties
│       ├── create_maxgraph_instance.sh
│       ├── data_loader.sh
│       ├── delete_maxgraph_instance.sh
│       ├── executor-entrypoint.sh
│       ├── executor.vineyard.properties
│       ├── expose_gremlin_server.sh
│       ├── frontend-entrypoint.sh
│       ├── frontend.vineyard.properties
│       ├── func.sh
│       ├── kill_process.sh
│       ├── load_data.sh
│       ├── load_modern_graph.sh
│       ├── manager-entrypoint.sh
│       ├── maven.settings.xml
│       ├── pod.yaml
│       └── set_config.sh
```

Similar to the files in distribution, most of the files are simple wrappers for a single-line Java command. In addition, it seems many files are for the same functions as those in `distribution`. (As the filename implies, starting coordinator/front-end and exposing an entrypoint.) 
It would be better if the scripts to start the coordinator (and for others) can be unified, whether to start it in a docker container/ or in a physical machine (or local). We may not need a set of docker-specific scripts and the `giectl` is sufficient. 

-------------------------------

I suggest we reorg them as:
```
interactive_engine
└── bin
    ├── giectl # with sub-commands, like `start-coordinator`, `start-frontend`, `stop-all` ....
    ├──data_loader.sh # data_loader can be separate if in need.
└── config
    ├── log4rs.yml 
    ├── logback.xml
    ├── frontend.vineyard.properties
    ├── executor.vineyard.properties
    ├── coordinator.application.properties
```

acezen:
## What do these changes do?
- graphscope-store log to both console and file
- failed the pod if store-cluster start failed
- move the store related command from giectl to store_ctl.

## Related issue number

Fixes #404 
Fixes #518


.github/workflows/ci.yml
```diff
@@ -630,15 +630,15 @@ jobs:
...
-        # helm uninstall ci
-        # helm install ci --set image.tag=${{ github.sha }},executor=${{ matrix.executor }} \
-        #                               ./graphscope-store
-        # helm test ci --timeout 5m0s
-        # python3 -m pytest -s vvv tests/kubernetes/test_store_service.py -k test_demo_loaded
+        helm uninstall ci
+        helm install ci --set image.tag=${{ github.sha }},executor=${{ matrix.executor }} \
+                                      ./graphscope-store
+        helm test ci --timeout 5m0s
+        python3 -m pytest -s vvv tests/kubernetes/test_store_service.py -k test_loaded_demo
```
siyuan0322: The name seems not intuitive... What about `test_demo_after_restart` and `test_demo_from_scratch`

tianliplus (approved):
LGTM

PR #777, opened 2021-09-07
clear duplicated and unused code

tianliplus:
## Related issue number

Fixes


yecol (approved):
Please double check the tags for the basic images, make sure that these changes are essential. 

I have no other comments! Thanks again for tidy up the codebase!

.github/workflows/ci.yml
```diff
@@ -440,7 +440,7 @@ jobs:
         fail_ci_if_error: true
 
     - name: Clean
-      if: always()
+      if: true
```
siyuan0322, line 443: Revert this before merging.

.github/workflows/graphscope-store.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.cn-hongkong.aliyuncs.com/graphscope/graphscope-runtime:latest
+FROM registry.cn-hongkong.aliyuncs.com/graphscope/graphscope-runtime:latest-root-user
```
siyuan0322, line 1: revert this before merging

tianliplus: done

.github/workflows/ci.yml
tianliplus: done

.github/workflows/ci.yml
```diff
@@ -592,6 +593,7 @@ jobs:
         name: gss-${{ github.sha }}
         path: |
           maxgraph.tar.gz
+          graphscope_store_data_load.tar.gz
```
siyuan0322, line 596: It would be better to add another `upload-artifact` step to only load the `graphscope_store_data_load.tar.gz`, or else it will be package together with maxgraph.tar.gz which size is much larger. It will be convenient for us to be able to download this artifact only.

tianliplus: done

PR #784, opened 2021-09-08
Fix graph store restart

### LINKED ISSUE #392: Fully test the restart store service on previous PersistentVolume.
Store service is designed for persistent storage, and should be able to work normally when restarted on previous volume with data.
This feature is not fully tested.

tianliplus:
## Related issue number

Fixes #392

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,10 +1,11 @@
-name: GraphScope CI
+name: GrphScope CI
 
 on:
   # Trigger the workflow on push or pull request, but only for the main branch
   push:
     branches:
       - main
+      - fix_restart
     paths-ignore:
       - 'gnn_engine/**'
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -679,13 +679,17 @@ jobs:
 
         # helm restart ci release and run demo with the PersistentVolume
         # FIXME(@haixia): restart seems not work.
```
siyuan0322, line 681: This FIXME can also be deleted.

siyuan0322 (approved):
LGTM.
The sleep here is treated as a temporal solution.
#785 is opened in order to fully resolve that.

.github/workflows/ci.yml
tianliplus: done

PR #793, opened 2021-09-09
`shutil.rmtree` cannot remove files

siyuan0322:
It will raise errors like `error: [Errno 20] Not a directory: '/root/GraphScope/coordinator/gscoordinator/builtin/app/builtin_app.zip'`,
Since there is `ignore_errors=True` these errors have been ignored before.

Also make the code more cleaner.

### DIFF
--- a/coordinator/setup.py
+++ b/coordinator/setup.py
@@ -16,10 +16,10 @@
 # limitations under the License.
 #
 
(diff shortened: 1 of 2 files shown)


sighingnow (approved):
Nice catch.

PR #814, opened 2021-09-15
Recover log4rs.yml of graphscope-store and fix the log output location

### LINKED ISSUE #813: Recover the log4rs setting of graphscope-store
- the log4rs.yml has been unified in [commit](https://github.com/alibaba/GraphScope/commit/032ed994a874525f55cf16f7bd05462cccbb491c). Since the `store-ctl` has been separate from `giectl`,  the log4rs.yml need to be recovered too.
- fix the log output location of `logback.yml` and `log4rs.yml`

acezen:
## What do these changes do?
- the log4rs.yml has been unified in [commit](https://github.com/alibaba/GraphScope/commit/032ed994a874525f55cf16f7bd05462cccbb491c). Since the `store-ctl` has been separate from `giectl`,  the log4rs.yml need to be recovered too.
- fix the log output location of `logback.yml` and `log4rs.yml`

## Related issue number

Fixes #813

### DIFF
--- a/interactive_engine/distribution/src/conf/log4rs.yml
+++ b/interactive_engine/distribution/src/conf/log4rs.yml
@@ -7,7 +7,7 @@ appenders:
   file:
     kind: rolling_file
     append: true
-    path: "${env:LOG_DIR:-/tmp/graphscope-store}/rust.log"
+    path: "/tmp/graphscope-store/rust.log"
     encoder:
       pattern: "{d(%Y-%m-%d %H:%M:%S.%f)} {h({l:<5})} (({f}:{L})) [{T}] {m}{n}"
     policy:
@@ -16,7 +16,7 @@ appenders:
         limit: 100mb
       roller:
         kind: fixed_window
-        pattern: "${env:LOG_DIR:-/tmp/graphscope-store}/rust.log.{}"
+        pattern: "/tmp/graphscope-store/rust.log.{}"
         count: 10
 
 root:
--- a/interactive_engine/distribution/src/conf/log4rs.yml
+++ b/interactive_engine/distribution/src/conf/log4rs.yml
@@ -7,7 +7,7 @@ appenders:
   file:
     kind: rolling_file
     append: true
-    path: "/tmp/graphscope-store/rust.log"
+    path: "/tmp/graphscope-store/store-executor.log"
     encoder:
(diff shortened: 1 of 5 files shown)


tianliplus (approved):

yecol:
Since in this PR, you are assigning log path, it raises two questions:
1. why do you assign two different log paths in one PR? i.e., `/tmp/logs` and `/tmp/graphscope-store`
2. in common practices, unix-like systems usually output log files to `/var/logs` or `/var/log` rather than `/tmp`, do we have any specific reasons to place logs in `/tmp`?

yecol:
It is OK to merge this PR. I open another issue regards to the comments above. 
#815

acezen:
> Since in this PR, you are assigning log path, it raises two questions:
> 
> 1. why do you assign two different log paths in one PR? i.e., `/tmp/logs` and `/tmp/graphscope-store`
> 2. in common practices, unix-like systems usually output log files to `/var/logs` or `/var/log` rather than `/tmp`, do we have any specific reasons to place logs in `/tmp`?

1. `/tmp/logs` is a default log4rs.yml setting of gie, `/tmp/graphscope-store` is for graphscope-store.
2. usr  `graphscope` has no permission to write in `/var`

yecol:
1. the "default" is not well designed then.
2. google it, use keywords like `log to var permission denied`, it is well answered.

interactive_engine/distribution/src/conf/log4rs.yml
```diff
@@ -1,109 +1,27 @@
...
     kind: console
     encoder:
       pattern: "{d(%Y-%m-%d %H:%M:%S.%f)} {h({l:<5})} (({f}:{L})) [{T}] {m}{n}"
-  executor:
-    kind: env_rolling_file
+  file:
+    kind: rolling_file
     append: true
-    path: "${env:LOG_DIRS:-./}/executor.log"
+    path: "/tmp/graphscope-store/rust.log"
```
yecol, line 10: why renames this filename?
I think `executor.log` is better than `rust.log`, it is more self-described.

acezen: I not sure if this is the log for executor ? @tianliplus Can you give a reply comment here ?

PR #855, opened 2021-09-21
Make the `add_edges` behavior more Pythonic

### LINKED ISSUE #727: The loading from edges behavior is diverged with the Doc.
After we have eager, the `sess.g()` will immediately be evaluated and a graph will be created in the backend.
But out previous design is one can only load from edge files when the graph is not created, otherwise we cannot determine whether it's an *only load from edges* operation or an *Error* that one forgets to specify the vertex labels.

So currently the loading from edges cannot be achieved using `g = sess.g(); g = g.add_edges()`, we can only fallback to the `load_from` style, but the doc gives no hints about that.

We should 
- make the document consistent, and 
- find a way to enable loading from edges in `g = sess.g()` style, which may be achieved by eager optimization by combining steps.

The eager and lazy interoperation is very subtle, we should be very careful about that and keep the document updated whenever something changed.

siyuan0322:
Now the `add_edges` will always deduce the vertex tables when not found in existed labels
Fixes #727

### DIFF
--- a/python/graphscope/framework/graph.py
+++ b/python/graphscope/framework/graph.py
@@ -428,9 +428,9 @@ class GraphDAGNode(DAGNode, GraphInterface):
     ):
         """Add edges to the graph, and return a new graph.
         Here the src_label and dst_label must be both specified or both unspecified,
-            i.   src_label and dst_label both unspecified and current graph has 0 vertex label.
+            i.   src_label and dst_label both unspecified and current graph has no vertex label.
                  We deduce vertex label from edge table, and set vertex label name to '_'.
-            ii.  src_label and dst_label both unspecified and current graph has 1 vertex label.
+            ii.  src_label and dst_label both unspecified and current graph has one vertex label.
                  We set src_label and dst label to this single vertex label.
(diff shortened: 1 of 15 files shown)


python/graphscope/framework/graph.py
```diff
@@ -420,34 +419,27 @@ def add_vertices(self, vertices, label="_", properties=None, vid_field=0):
...
     ):
         """Add edges to the graph, and return a new graph.
-
-        1. Add edges to a uninitialized graph.
-
-            i.   src_label and dst_label both unspecified. In this case, current graph must
-                 has 0 (we deduce vertex label from edge table, and set vertex label name to '_'),
-                 or 1 vertex label (we set src_label and dst label to this).
+        Here the src_label and dst_label must be both specified or both unspecified,
+            i.   src_label and dst_label both unspecified and current graph has 0 vertex label.
```
sighingnow, line 431: `has 0 vertex label` -> `has no vertex label`.

sighingnow (approved):
LGTM.

sighingnow:
You may need to update the required version in CMakeLists.txt as well.

PR #945, opened 2021-11-04
Add backup e2e test

### LINKED ISSUE #946: Add e2e test for full link backup module
as titled.

GoldenLeaves:
As titled.

Fixes #946

### DIFF
--- a/interactive_engine/sdk/backup_test.sh
+++ b/interactive_engine/sdk/backup_test.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+set -x
+base_dir=$(cd `dirname $0`; pwd)
+ps -ef | grep "com.alibaba.maxgraph.servers.MaxNode" | grep -v grep | awk '{print $2}' | xargs kill -9
+cd ${base_dir}/.. && mvn clean install -DskipTests -Pv2
+cd ./distribution/target/ && tar xvzf maxgraph.tar.gz && cd maxgraph
+
+# start server
+maxgraph_dir=$(pwd)
+sed -e "s@LOG4RS_CONFIG@${maxgraph_dir}/conf/log4rs.yml@g" \
+    -e "s@partition.count=8@partition.count=4@g" \
+    -e "s@backup.enable=false@backup.enable=true@g" \
+    -e "s@log.recycle.enable=true@log.recycle.enable=false@g" \
+    conf/config.template > /tmp/max_node_backup_test.config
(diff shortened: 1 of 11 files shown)


interactive_engine/sdk/backup_test.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+set -x
+base_dir=$(cd "$(dirname "$0")"; pwd)
+ps -ef | grep "com.alibaba.maxgraph.servers.MaxNode" | grep -v grep | awk '{print $2}' | xargs kill -9
+cd ${base_dir}/.. && mvn clean install -DskipTests -Pv2
+cd ./distribution/target/ && tar xvzf maxgraph.tar.gz && cd maxgraph
```
siyuan0322, line 6: I suggest split the `install` stage and `test` stage into two steps in local.yml.

interactive_engine/sdk/backup_test.sh
```diff
@@ -0,0 +1,27 @@
...
+maxgraph_dir=$(pwd)
+sed -e "s@LOG4RS_CONFIG@${maxgraph_dir}/conf/log4rs.yml@g" \
+    -e "s@partition.count=8@partition.count=4@g" \
+    -e "s@backup.enable=false@backup.enable=true@g" \
+    -e "s@log.recycle.enable=true@log.recycle.enable=false@g" \
+    conf/config.template > /tmp/max_node_backup_test.config
+LOG_NAME=maxnode MAXGRAPH_CONF_FILE=/tmp/max_node_backup_test.config ./bin/store_ctl.sh max_node_maxgraph &
+sleep 20
+# load data
+cd ${base_dir} && mvn -Dtest=com.alibaba.graphscope.groot.sdk.ClientBackupTest test
```
siyuan0322, line 18: If you could put those commands to start servers, and `mvn test` directly into the local.yml, then you can get rid of the extra checking of `exit_code`, since if the code doesn't equal to 0, it will fail the CI, and then there is no need to clean up the process or data.

interactive_engine/sdk/backup_test.sh
GoldenLeaves: done

interactive_engine/sdk/backup_test.sh
GoldenLeaves: What I think is to write like this is more convenient to test in my own local

GoldenLeaves: And the impact of this piece on ci should be very very small

siyuan0322: OK~

tianliplus (approved):

PR #994, opened 2021-11-15
[Bug Fix] fix nullptr in arrow_projected_fragment.h

septicmk:
## What do these changes do?
Arrow_projected_fragment causes a bug. When `edge_prop_` is -1. (A graph with no edge property).

https://github.com/alibaba/GraphScope/blob/699eea6cc187bdfaf9d5371ae5c5ba7b25963ffc/analytical_engine/core/fragment/arrow_projected_fragment.h#L573-L577
will set `edge_data_array_` to `nullptr`(correct)

https://github.com/alibaba/GraphScope/blob/699eea6cc187bdfaf9d5371ae5c5ba7b25963ffc/analytical_engine/core/fragment/arrow_projected_fragment.h#L686-L688
Unfortunately, a general method `GetEdgeNum()` will reference `edge_data_array_` without check.


sighingnow (approved):
LGTM.

PR #1030, opened 2021-11-22
Add make install support for lgraph

GoldenLeaves:
## What do these changes do?

Add make install module for lgraph repo.


interactive_engine/lgraph/src/log_subscription/subscriber.cc
```diff
@@ -36,9 +35,6 @@ Subscriber::~Subscriber() {
 
 LogMessage Subscriber::Poll(size_t timeout_ms) {
   auto kafka_msg = consumer_.poll(std::chrono::milliseconds(timeout_ms));
-  if (kafka_msg && !kafka_msg.get_error()) {
-    current_offset_ = kafka_msg.get_offset();
-  }
   return LogMessage{std::move(kafka_msg)};
```
LiSu, line 38: Why is this check removed? Do we assume that every poll action will succeed?

LiSu (approved):
LGTM

PR #1040, opened 2021-11-23
check no duplicate property name

### LINKED ISSUE #1022: [BUG] Ensure that a property name would not appear more than twice within a label.
**Describe the bug**

GIE requires that property names are unique in every label, and now the loading module doesn't force that. But it could happen if user put the id field in properties, or the source file have two columns having the same name.

An additional sanity check should be performed after the arrow Table is created.

**Additional context**
Orignally reported in #973, also thanks @longbinlai and @sighingnow for their investigation.

siyuan0322:
## Related issue number

Fixes #1022


sighingnow (approved):
LGTM.

PR #1061, opened 2021-11-26
Add some functions for starting Pegasus in GraphScope

BingqingLyu:
## Related issue number

Fixes

### DIFF
--- a/research/engine/pegasus/network/src/manager.rs
+++ b/research/engine/pegasus/network/src/manager.rs
@@ -13,13 +13,13 @@
 //! See the License for the specific language governing permissions and
 //! limitations under the License.
 
-use std::net::{SocketAddr, ToSocketAddrs};
-
 use crate::config::ConnectionParams;
 use crate::{NetError, Server};
+use std::net::{SocketAddr, ToSocketAddrs};
+use std::sync::{Arc, Mutex};
 
 pub trait ServerDetect: Send {
-    fn fetch(&mut self) -> &[Server];
+    fn fetch(&self) -> Vec<Server>;
 }
 
 #[allow(dead_code)]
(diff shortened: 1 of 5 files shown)


research/engine/pegasus/pegasus/src/lib.rs
```diff
@@ -324,6 +327,7 @@ fn allocate_local_worker(conf: &Arc<JobConf>) -> Result<Option<WorkerIdIter>, Bu
 }
 
 use std::sync::Once;
+use std::net::SocketAddr;
```
longbinlai, line 330: Move these `use` to the head of the file

research/engine/pegasus/pegasus/src/lib.rs
```diff
@@ -183,13 +183,16 @@ pub fn startup_with<D: ServerDetect + 'static>(conf: Configuration, detect: D) -
...
         let conn_conf = net_conf.get_connection_param();
         let addr = pegasus_network::start_up(server_id, conn_conf, addr, detect)?;
         info!("server {} start on {:?}", server_id, addr);
-    }
-    Ok(())
+        Some(addr)
+    } else {
+        None
+    };
+    Ok(res)
```
longbinlai, line 195: Directly return `Ok(if let Some(...) {} else { ... })`, instead of `let res = if let ...; Ok(res)`

longbinlai (commented):
Small fix.

research/engine/pegasus/pegasus/src/lib.rs
BingqingLyu: Done

research/engine/pegasus/pegasus/src/lib.rs
BingqingLyu: Done

research/engine/pegasus/network/src/manager.rs
```diff
@@ -58,7 +58,36 @@ impl ServerManager {
...
+}
+
+impl SimpleServerDetector {
+    pub fn new() -> Self {
+        SimpleServerDetector { peers_mutex: Mutex::new(vec![]) }
+    }
+
+    pub fn update_peer_view<Iter: Iterator<Item = (u64, SocketAddr)>>(&self, peer_view: Iter) {
+        let new_peers = peer_view.map(|(id, addr)| Server { id, addr }).collect::<Vec<Server>>();
+        let mut peers = self.peers_mutex.lock().expect("unexpected error locking when update peer view");
```
longbinlai, line 77: Will this `expect()` cause any issue? I mean if it is used while running a service, then it should not be down due to such an error; if it is just for initiating a service, then it is fine.

research/engine/pegasus/network/src/manager.rs
```diff
@@ -58,7 +58,36 @@ impl ServerManager {
...
+        *peers = new_peers;
+    }
+}
+
+impl ServerDetect for SimpleServerDetector {
+    fn fetch(&self) -> Vec<Server> {
+        let peers = self.peers_mutex.lock().expect("unexpected error locking when fetch servers");
```
longbinlai, line 84: and this `expect()`.

research/engine/pegasus/network/src/manager.rs
BingqingLyu: This is just for initiating the service (used when starting service with v2 store).

PR #1067, opened 2021-11-29
Fixes failure of nightly CI

lidongze0629:
## Related issue number

Fixes

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -187,7 +187,7 @@ def build_learning_engine():
     extra_compile_args = []
     extra_link_args = []
 
-    include_dirs.append("/usr/local")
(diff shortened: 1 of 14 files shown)


yecol (approved):

python/setup.py
```diff
@@ -187,6 +187,7 @@ def build_learning_engine():
     extra_compile_args = []
     extra_link_args = []
 
+    include_dirs.append("/usr/local")
```
sighingnow, line 190: Include dir is `/usr/local/include`, rather than `/usr/local`.

sighingnow (changes requested):

sighingnow (approved):

PR #1070, opened 2021-11-29
Implemented SortLimit and SortLimitBy.

luoxiaojian:


research/engine/pegasus/pegasus/src/operator/concise/limit.rs
```diff
@@ -60,3 +66,230 @@ impl<D: Data> Limit<D> for Stream<D> {
...
+        self.sort_limit_by(size, |x, y| x.cmp(y))
+    }
+}
+
+impl<D: Data> SortLimitBy<D> for Stream<D> {
+    fn sort_limit_by<F>(self, size: u32, cmp: F) -> Result<Stream<D>, BuildJobError> where F: Fn(&D, &D) -> Ordering + Send + 'static {
+        let share_cmp = ShadeCmp { cmp: Rc::new(cmp) };
+        let cmp_clone = ShadeCmp { cmp: share_cmp.cmp.clone() };
+
+        let local_sort = sort_limit_by_partition(self, size, cmp_clone, false)?;
```
longbinlai, line 81: Better define `sort_limit_by_partition(.., name: String, )`, and we can give the two calls different names.
One with `sort_limit_by_partition(..., "sort_limit_by_partition_locally")`
The other with `sort_limit_by_partition(..., "sort_limit_by_partition_globally")`

research/engine/pegasus/pegasus/src/operator/concise/limit.rs
```diff
@@ -60,3 +66,230 @@ impl<D: Data> Limit<D> for Stream<D> {
...
+                    Ok(())
+                })
+            }
+        })
+    }
+}
+
+type Cmp<D> = Rc<dyn Fn(&D, &D) -> Ordering + Send + 'static>;
+
+struct ShadeCmp<C> {
```
longbinlai, line 163: Not sure why we need `ShadeCmp`? `Cmp<D>` alone should do all the work.

research/engine/pegasus/pegasus/src/operator/concise/limit.rs
```diff
@@ -60,3 +66,230 @@ impl<D: Data> Limit<D> for Stream<D> {
...
+    pub cmp: Rc<C>,
+}
+
+unsafe impl<C: Send> Send for ShadeCmp<C> {}
+
+struct FixedSizeHeap<D> {
```
longbinlai, line 169: We may want to use `BinaryHeap` from `std`. Note that the standard `BinaryHeap` requires the data to be `Ord`, which may not be the case for our `D` in `sort_limit_by`. For that we may implement a structure like:
```
struct DataWithCmp<D, C> {
    data: D,
    cmp: Cmp<C>
}

impl Ord for DataWithCmp ... 
```

longbinlai (commented):
Using `BinaryHeap` from `std::collections` may be a better way.

longbinlai (commented):
Small fix.

PR #1071, opened 2021-11-29
Give appropriate labels to GraphScope in Kubernetes

### LINKED ISSUE #938: Give appropriate labels to graphscope when deployed in kubernetes.
Assign appropriate labels to pods/services/deployments of graphscope can make it much easier to delete deployed graphscope, as labels can group components in one cluster together. or we need to delete those components one by one (When deployed by Python) like:
```shell
# Suppose the name is xxx
kubectl delete deployments gs-engine-xxx
kubectl delete pods gs-etcd-xxx
```
Instead it's better to use labels to manage them, like:
```shell
kubectl delete pods -l app.kubernetes.io/instance=xxx
```
 And it's more aligned with the best practice, make room for future management, rollbacks, updates, scale, etc.

And we can add this feature to Helm charts as well.
- [ ] Deployed by Python
- [x] Deployed by Helm 

Reference: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

lidongze0629:
## Related issue number

Fixes #938

### DIFF
--- a/coordinator/gscoordinator/cluster.py
+++ b/coordinator/gscoordinator/cluster.py
@@ -66,6 +66,7 @@ from gscoordinator.utils import INTERACTIVE_ENGINE_SCRIPT
 from gscoordinator.utils import WORKSPACE
 from gscoordinator.utils import ResolveMPICmdPrefix
 from gscoordinator.utils import parse_as_glog_level
+from gscoordinator.version import __version__
 
 logger = logging.getLogger("graphscope")
 
@@ -418,7 +419,27 @@ class KubernetesClusterLauncher(Launcher):
 
     def _create_mars_scheduler(self):
         logger.info("Launching mars scheduler pod for GraphScope ...")
-        labels = {"name": self._mars_scheduler_name}
+
+        labels = {
+            "app.kubernetes.io/name": self._mars_scheduler_name,
+            "app.kubernetes.io/instance": self._instance_id,
+            "graphscope.components": "mars",
+            "graphscope.version": __version__,
+        }
+
+        # create mars service
+        service_builder = ServiceBuilder(
+            self._mars_service_name,
+            service_type=self._saved_locals["service_type"],
+            port=self._mars_scheduler_port,
+            selector=labels,
+        )
+        self._resource_object.append(
+            self._core_api.create_namespaced_service(
+                self._saved_locals["namespace"], service_builder.build()
+            )
+        )
+
         # create engine replicaset
         scheduler_builder = self._gs_mars_scheduler_builder_cls(
             name=self._mars_scheduler_name,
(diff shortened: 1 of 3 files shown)


coordinator/gscoordinator/cluster.py
```diff
@@ -418,7 +419,27 @@ def close_interactive_instance(self, object_id):
 
     def _create_mars_scheduler(self):
         logger.info("Launching mars scheduler pod for GraphScope ...")
-        labels = {"name": self._mars_scheduler_name}
+
+        labels = {
+            "app.kubernetes.io/name": self._mars_scheduler_name,
+            "app.kubernetes.io/instance": self._instance_id,
+            "graphscope.components": "mars",
```
yecol, line 426: it is not appropriate to include mars as graphscope.components.

coordinator/gscoordinator/cluster.py
```diff
@@ -418,7 +419,27 @@ def close_interactive_instance(self, object_id):
 
     def _create_mars_scheduler(self):
         logger.info("Launching mars scheduler pod for GraphScope ...")
-        labels = {"name": self._mars_scheduler_name}
+
+        labels = {
+            "app.kubernetes.io/name": self._mars_scheduler_name,
+            "app.kubernetes.io/instance": self._instance_id,
+            "graphscope.externals": "mars",
```
sighingnow, line 426: `. externals` ?

yecol: @sighingnow give a suggestion?

sighingnow: How about `.component`?

siyuan0322 (commented):
> app.kubernetes.io/name | The name of the application | mysql | string

I suggest the `app.kubernetes.io/name` always equal to `graphscope`.
Maybe also add a `version` field?
And we use `instance` to distinguish difference graphscope clusters.
I think the `engine_name` doesn't need to appear in labels.

Reference: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels

lidongze0629:
> > app.kubernetes.io/name | The name of the application | mysql | string
> 
> I suggest the `app.kubernetes.io/name` always equal to `graphscope`. Maybe also add a `version` field? And we use `instance` to distinguish difference graphscope clusters. I think the `engine_name` doesn't need to appear in labels.
> 
> Reference: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels

Thanks for your recommended labels. I will update labels to
```
app.kubernetes.io/name: graphscope
app.kubernetes.io/instance: unique_instance_id
app.kubernetes.io/version: "0.8.0"
app.kubernetes.io/component: engine/coordinator/etcd
app.kubernetes.io/component-name: engine_name/coordinator_name/etcd_name
```
`engine_name` is must for service selector

siyuan0322:
> unique_instance_id

What about using the label `instance` together with `component`, which can select the proper pods and you won't need to update the engine name. This can be useful for our future scaling scenario, which can automatically include the newly added pod in.

lidongze0629:
> What about using the label `instance` together with `component`, which can select the proper pods and you won't need to update the engine name. This can be useful for our future scaling scenario, which can automatically include the newly added pod in.

I think the labels above is enough for selector.

siyuan0322:
> > What about using the label `instance` together with `component`, which can select the proper pods and you won't need to update the engine name. This can be useful for our future scaling scenario, which can automatically include the newly added pod in.
> 
> I think the labels above is enough for selector.

OK~ I was replied to the `engine-name` part, which is now gone.

sighingnow:
> I suggest the `app.kubernetes.io/name` always equal to `graphscope`.

+1

sighingnow (approved):

PR #1092, opened 2021-12-02
[GAIA/ENGINE] revert NetworkConfig to an old version

BingqingLyu:
## What do these changes do?

Revert NetworkConfig to an old version since the new version seems to be not compatible with startup_with ServerDetect.

## Related issue number

Fixes

### DIFF
--- a/research/engine/pegasus/network/src/config.rs
+++ b/research/engine/pegasus/network/src/config.rs
@@ -135,6 +135,8 @@ impl ConnectionParams {
 #[derive(Debug, Deserialize)]
 pub struct NetworkConfig {
     pub server_id: u64,
+    ip: String,
+    port: u16,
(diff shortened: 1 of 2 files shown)


research/engine/pegasus/network/src/config.rs
```diff
@@ -147,13 +149,15 @@ pub struct NetworkConfig {
...
 #[derive(Debug, Deserialize)]
 pub struct ServerConfig {
+    server_id: u64,
     ip: String,
     port: u16,
 }
 
 impl ServerConfig {
-    pub fn new(ip: String, port: u16) -> Self {
+    pub fn new(server_id:u64, ip: String, port: u16) -> Self {
```
longbinlai, line 158: format issue `server_id: u64`

PR #1104, opened 2021-12-03
[Gaia] Update gaia README

longbinlai:
As titled.


yecol (approved):
Mergeable, 
I have a suggestion: why not place this README in the root of `/research`~

longbinlai:
> Mergeable, I have a suggestion: why not place this README in the root of `/research`~

Okay, but probably `research/query_service`

research/query_service/gremlin/scripts/run_gaia.py
```diff
@@ -0,0 +1,247 @@
...
+                      args.raw_partitions, args.workers, num_machines, i, host_list_file)
+            cmd = ["ssh", hosts[i][0], run_params]
+            print(cmd)
+            pool.append(subprocess.Popen(cmd))
+        sync_process(pool)
+    else:
+        cmd = "%s %s %s %s %s %d -w %d" % (env, program, args.hadoop_home, args.raw_data_dir, tmp_dir,
+                      args.raw_partitions, args.workers)
+        print(cmd)
+        os.system(cmd)
```
lidongze0629, line 130: you can replace `os.system` with:

```py
subprocess.check_call([cmd])
```

longbinlai: Hi, without `shell=True`, the script cannot run. Is there a solution
Screen Shot 2021-12-03 at 14 33 43
?

research/query_service/gremlin/scripts/run_gaia.py
```diff
@@ -0,0 +1,247 @@
...
+            cmd = ["ssh", hosts[i][0], run_params]
+            print(cmd)
+            pool.append(subprocess.Popen(cmd, stdout=log_file, stderr=log_file))
+            log_file.close()
+        sync_process(pool)
+    else:
+        cmd = "%s %s %s %s %d -w %d" % \
+                     (env, program, local_dir, args.graph_dir, args.raw_partitions, args.workers)
+        print(cmd)
+        os.system(cmd)
```
lidongze0629: ditto

research/query_service/gremlin/scripts/run_gaia.py
```diff
@@ -0,0 +1,247 @@
...
+            cmd = ["ssh", hosts[i][0], run_params]
+            print(cmd)
+            pool.append(subprocess.Popen(cmd, stdout=log_file, stderr=log_file))
+            prev_host = hosts[i][0]
+            log_file.close()
+        sync_process(pool)
+    else:
+        cmd = "%s %s" % (env, program)
+        print(cmd)
+        os.system(cmd)
```
lidongze0629: ditto

research/query_service/gremlin/scripts/utils.py
```diff
@@ -0,0 +1,107 @@
...
+
+def str2bool(v):
+    """
+    Parse a string option either of yes/true/t/y/1 into `True`, and a a string of no/false/f/n/0 into `False`
+    :param v: The string option
+    :return: A boolean value from the string
+    """
+    if isinstance(v, bool):
+        return v
+    if v.lower() in ('yes', 'true', 't', 'y', '1'):
```
lidongze0629, line 102:
```python
if v.lower() in ('yes', 'true', 't', 'y', '1'):
    return True
if v.lower() in ('no', 'false', 'f', 'n', '0'):
    return False

raise argparse.ArgumentTypeError('Boolean value expected.')
```

research/query_service/gremlin/scripts/run_gaia.py
lidongze0629: From [Doc](https://docs.python.org/3/library/subprocess.html#subprocess.check_call):

> check_call: Run command with arguments. Wait for command to complete. If the return code was zero then return, otherwise raise CalledProcessError. The CalledProcessError object will have the return code in the returncode attribute. If check_call() was unable to start the process it will propagate the exception that was raised.

```python
subprocess.check_call(["python3", "/home/admin/xxx.py", "arg1", "arg2"], env=dict(os.environ, PATH=PATH))
```

research/query_service/gremlin/scripts/run_gaia.py
```diff
@@ -0,0 +1,246 @@
...
+        return None, None
+    else:
+        return absolute_path, filename
+
+
+####################################################################
+####################### Prepare the arguments ######################
+####################################################################
+
+TMP_FOLDER = "tmp4p1wlx9p"
```
sighingnow, line 49: Use `with tempfile.TemporaryFile(): ...` instead.

longbinlai: This is fine. As the folder is a temporary one.  I change the file name to "gaia_tmp" to avoid mis-understanding

research/query_service/gremlin/scripts/run_gaia.py
longbinlai: I removed these codes. But there are other issues like "partial exectuble path."

lidongze0629: Address the problem with:

```python
cmd = ["cmake", "."]  # not ok

cmd = [shutil.which("cmake"), "."] # is ok
```

longbinlai: Nice, it is okay now. Thank you @lidongze0629

PR #1129, opened 2021-12-08
Allow feed a `context_type` when loading custom gars.

sighingnow:
## What do these changes do?

It is quite strange that builtin gars accept a `context` argument for context type, but when loading custom gars it is not allowed.

When there's a `context_type` entry in the gar config, it would still be overwrite, but we need to check if the value is available first.

## Related issue number

N/A


sighingnow:
Prompted by @septicmk

yecol (approved):

PR #1131, opened 2021-12-08
Fix failure of dag.add_op during running app_assets twice

lidongze0629:
## Related issue number

Fixes

### DIFF
--- a/python/graphscope/framework/dag.py
+++ b/python/graphscope/framework/dag.py
@@ -46,6 +46,11 @@ class Dag(object):
     def __repr__(self):
         return self.__str__()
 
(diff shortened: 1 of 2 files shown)


python/graphscope/framework/dag.py
```diff
@@ -46,6 +46,11 @@ def __str__(self):
     def __repr__(self):
         return self.__str__()
 
+    def exist(self, op):
```
siyuan0322, line 49: exists

siyuan0322 (approved):

septicmk (approved):

PR #1156, opened 2021-12-13
Charts: Fix the GraphScope store mount volume permission issue

haoxins:
I think this could fix the below problems:

1. `statefulset.apps/graphscope-store-store`

```
Caused by: java.nio.file.AccessDeniedException: /var/lib/graphscope-store/8
```

2. `statefulset.apps/graphscope-store-coordinator`

```
Caused by: java.nio.file.AccessDeniedException: /etc/graphscope-store/my.meta/query_snapshot_info
```

### DIFF
--- a/charts/graphscope-store/templates/coordinator/statefulset.yaml
+++ b/charts/graphscope-store/templates/coordinator/statefulset.yaml
@@ -129,6 +129,12 @@ spec:
             - name: config
               mountPath: /etc/graphscope-store/my.cnf.tpl
               subPath: my.cnf
+
+      securityContext:
+        runAsUser: 1001
+        runAsGroup: 1001
+        fsGroup: 1001
+
       volumes:
         - name: config
           configMap:
--- a/charts/graphscope-store/templates/store/statefulset.yaml
+++ b/charts/graphscope-store/templates/store/statefulset.yaml
@@ -143,6 +143,11 @@ spec:
               mountPath: /etc/graphscope-store/my.cnf.tpl
               subPath: my.cnf
 
+      securityContext:
+        runAsUser: 1001
+        runAsGroup: 1001
+        fsGroup: 1001
+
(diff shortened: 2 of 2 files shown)


siyuan0322:
May I know the context of this exception, like the machine environment and steps to reproduce it? Since I've never seen that.

The CI failed and I'm rerunning it to see if it's caused by the ﬂakiness.

However the default user `graphscope` in our image has the following id:
```shell
[graphscope@127beddcb699 ~]$ id
uid=1001(graphscope) gid=1001(graphscope) groups=1001(graphscope)
```

Therefore the value of the `securityContext` might need to be configured to `1001`?

haoxins:
When I install the Helm charts in Kubernetes `v1.21`, not sure this is the root cause.

haoxins:
> The CI failed and I'm rerunning it to see if it's caused by the ﬂakiness.

No, I found some mistakes, will fix 😂

siyuan0322:
I made some experiment on the version you mentioned and confirmed that the value of securityContext should be set as:
```
securityContext:
  runAsUser: 1001
  runAsGroup: 1001
  fsGroup: 1001
```
So as to not conflict with the user in the image. provided in the [dockerfile](https://github.com/alibaba/GraphScope/blob/4090264099602ef2721542d300b3ed1d4c39641a/k8s/manylinux2014.Dockerfile#L329) we have already set the user `graphscope` whose id is 1001.

I think this would probably address the failure of the CI and probably your problem, but I can't reproduce it now.

Could you change the securityContext in this PR so that we can get it landed after the CI turned green, and we can look further to your permission problem if it still happends.

haoxins:
Yeah, you are right~

I just post the results from my container logs.
And all my deployments are green now.

siyuan0322 (approved):
LGTM, thanks!

PR #1161, opened 2021-12-14
chore - Add default container annotation to engine replica set

haoxins:


siyuan0322 (approved):
Overall LGTM.
However there's an exception here:
```
Error: -14 17:24:32,720 [ERROR][cluster:1170]: Error when launching GraphScope on kubernetes cluster: type object 'super' has no attribute 'add_annotation'
```

python/graphscope/deploy/kubernetes/resource_builder.py
```diff
@@ -799,6 +804,7 @@ def add_engine_container(self, name, image, cpu, mem, preemptive, **kwargs):
                 }
             )
         )
+        super.add_annotation("kubectl.kubernetes.io/default-container", name)
```
siyuan0322, line 807: This should be `super().add_annotation(`

siyuan0322 (changes requested):

python/graphscope/deploy/kubernetes/resource_builder.py
haoxins: Sry, fixed~

siyuan0322 (approved):

PR #1188, opened 2021-12-17
Support setting ports of coordinator pod from GSCOORDINATOR_PORTS env

lidongze0629:
## What do these changes do?

Sometimes, support to open multiple ports coordinator pod is needed, such as launching an `sshd` service, or creating GIE gremlin frontend service with `HostNetwork=true`.

## Related issue number

Fixes


sighingnow (approved):

yecol (approved):
It would be better to add 1-2 sentences in the summary describing the context,  especially a PR has no related issues.

PR #1247, opened 2021-12-31
Fixes the long-standing "Event loop is closed" issue.

sighingnow:
## What do these changes do?

The `DriverRemoteConnection` is leaked.

The failure can be reproduced when `traverse_source` involved.

## Related issue number

N/A


sighingnow (commented):
The `Try, Except, Pass` pattern is OK here.

siyuan0322 (approved):

PR #1268, opened 2022-01-12
Drop support for Python 3.6

### LINKED ISSUE #1269: Drop support for python 3.6
since the NetworkX version we depends on has been bump up to 2.6 and it only provides wheel packages for Python 3.7 - 3.9, the GraphScope should drop support for python 3.6 to consistent with NetworkX.

acezen:
## What do these changes do?

since the NetworkX dependency has been bump up to 2.6 and it has droped support for Python3.6,  the GraphScope should drop support for python 3.6.

## Related issue number

Fixes #1269

### DIFF
--- a/.github/workflows/build-graphscope-wheels-linux.yml
+++ b/.github/workflows/build-graphscope-wheels-linux.yml
@@ -203,26 +203,26 @@ jobs:
         tar xzf Python-3.7.9.tgz
         cd Python-3.7.9 && ./configure --enable-optimizations --enable-shared
         make altinstall
-        python3.7 -m pip install --upgrade pip --user
+        python3 -m pip install --upgrade pip --user
 
         # install graphscope-client
         cd ${GITHUB_WORKSPACE}/artifacts
         tar -zxf ./wheel-${{ github.sha }}/client.tar.gz
         pushd python/dist/wheelhouse
-        for f in * ; do python3.7 -m pip install $f --user || true; done
+        for f in * ; do python3 -m pip install $f --user || true; done
         popd
 
         # install graphscope
         tar -zxf ./wheel-${{ github.sha }}/graphscope.tar.gz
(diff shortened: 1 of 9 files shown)


.github/workflows/build-graphscope-wheels-linux.yml
```diff
@@ -193,10 +196,13 @@ jobs:
...
-        yum install -y gcc gcc-c++ make cmake
+        dnf install -y gcc gcc-c++ openssl-devel bzip2-devel libffi-devel wget make cmake git
 
-        # install git and python3
-        yum install -y git python3-pip python3-devel
+        # install python3.7
+        cd /tmp && wget https://www.python.org/ftp/python/3.7.9/Python-3.7.9.tgz
+        tar xzf Python-3.7.9.tgz
+        cd Python-3.7.9 && ./configure --enable-optimizations --enable-shared
+        make install
```
sighingnow, line 205: Could we install python 3.8 (I think it should be possible, as https://forums.centos.org/viewtopic.php?p=319600) using dnf or yum instead?

Building a custimized python for every CI runs seems not a good idea.

acezen: done~

lidongze0629 (approved):

python/graphscope/__init__.py
```diff
@@ -16,6 +16,13 @@
 # limitations under the License.
 #
 
+import sys
+
+if sys.version_info[:2] < (3, 7):
+    m = "Python 3.7 or later is required for GraphScope (%d.%d detected)."
+    raise ImportError(m % sys.version_info[:2])
+del sys
```
sighingnow, line 24: No need for that. We won't distribute wheels that built with python 3.6 anymore so that users have no change to installed the wheels under python 3.6.

Issue #1270, opened 2022-01-13
[networkx] Remaining optimizations for engine processing edges/nodes_to_modified


acezen:
**Is your feature request related to a problem? Please describe.**
[pull request](https://github.com/alibaba/GraphScope/pull/1250) has apply optimization for edges/nodes processing of client, but from [benchmark profiling](https://github.com/alibaba/GraphScope/issues/999), there still has some overheads on  engine receive  the op and process the
- engine parse the json 
- convert oid to gid and add new vertex to vertex_map (vertex-map related)
- make a benchmark, seems `rapidjson::Value` can  replace `folly::Dynamic`;

**Describe the solution you'd like**
for json parse, we could use RapidJson to replace folly::Json
- [ ] replace with rapid-json

**Additional context**
#999

PR #1271, opened 2022-01-13
 Replace `folly::dynamic` with `rapidjson::Value` and bump vineyard to 0.3.18

acezen:
## What do these changes do?
Since the `folly::dynamic` parse performance and construct object  performance are too bad.  After some benchmarking, we decide to replace `folly::dynamic` with `rapidjson::Value`. Below are some changes details.
-  Add a `dynamic::Value` class based on `rapidjson::Value` which support more friendly APIs.
- Replace the `folly::dynamic` with `dynamic::Value` and pass the test cases.

## Related issue number

Fixes https://github.com/alibaba/GraphScope/issues/1270

## To do:
- [x] new runtime image
- [x] vertex map for `dynamic::Value`


sighingnow:
Could we get rid of `folly::dynamic` and use a `rapidjson::Value` instead?

acezen:
> Could we get rid of `folly::dynamic` and use a `rapidjson::Value` instead?

Yes we can,  actually I'm doing a tiny benchmark between folly::Dynamic and rapidJson.

analytical_engine/apps/boundary/edge_boundary.h
```diff
@@ -53,16 +53,18 @@ class EdgeBoundary : public AppBase<FRAG_T, EdgeBoundaryContext<FRAG_T>>,
     vid_t gid;
     vertex_t u;
     bool no_target_set = true;
-    for (const auto& node : source_array.GetArray()) {
-      if (frag.Oid2Gid(dynamic_to_oid<oid_t>(node), gid)) {
+    for (auto& node : source_array) {
+      dynamic::Value oid(node);
+      if (frag.Oid2Gid(dynamic_to_oid<oid_t>(oid), gid)) {
```
sighingnow, line 58: Looks strange about the `dynamic::Value oid(node)`, why we need such a (implicit) copy as we will obtain a vanilla `oid_t` value and the reference won't be used any more?

acezen: Because in `for(auto& node: source_array)` , the node is `rapidjson::Value`, not `dynamic::Value`.

sighingnow: Could we have a overload for `dynamic_to_oid<oid_t>(rapidjson::Value const &)` as well?

analytical_engine/core/fragment/dynamic_fragment.h
```diff
@@ -2740,7 +2736,7 @@ class DynamicFragment {
     ivdata_.clear();
     ivdata_.resize(other->ivdata_.size());
     for (size_t i = 0; i < ivnum_; ++i) {
-      ivdata_[i] = other->ivdata_[i];
+      ivdata_[i].CopyFrom(other->ivdata_[i]);
```
acezen, line 2739:
```suggestion
      ivdata_[i] = other->ivdata_[i];
```

analytical_engine/core/fragment/dynamic_fragment.h
```diff
@@ -2947,17 +2943,17 @@ class DynamicFragment {
...
     vid_t gid, dst_gid;
-    for (auto& oid : induced_vertices) {
+    for (const auto& oid : induced_vertices) {
       if (origin->GetVertex(oid, vertex)) {
         if (origin->IsInnerVertex(vertex)) {
           // store the vertex data
           CHECK(vm_ptr_->GetGid(fid_, oid, gid));
           auto lid = iv_gid_to_lid(gid);
-          ivdata_[lid] = origin->GetData(vertex);
+          ivdata_[lid].CopyFrom(origin->GetData(vertex));
```
acezen, line 2956:
```suggestion
          ivdata_[lid] = origin->GetData(vertex);
```

analytical_engine/apps/boundary/edge_boundary.h
acezen: Fixed.

analytical_engine/core/object/dynamic.cc
```diff
@@ -0,0 +1,59 @@
...
+
+/* static definition */
+gs::dynamic::AllocatorT gs::dynamic::Value::allocator_;
+
+std::size_t gs::dynamic::Value::hash() const {
+  switch (GetType()) {
+  case rapidjson::kNullType:
+    return 0xBAAAAAAD;
+  case rapidjson::kObjectType: {
+    return 0xBAAAAAAE;
```
sighingnow, line 31: If all node that has oid type `kObjectType` (i.e., json object), the hash value would be the same and they will be routed into the same fragment.

That sounds not the desired behaviour.

acezen: Change to not allow object as node( like NetworkX do), if the value is object, throw an runtime error.

sighingnow (approved):
LGTM.

lidongze0629 (approved):

PR #1310, opened 2022-02-08
Use boost::lexical_cast to handle string id fragment

siyuan0322:
`std::to_string` doesn't have the `std::to_string(std::string)` overload, thus when used with string id fragment, it will error out.


sighingnow (approved):

lidongze0629 (approved):

acezen:
Seems NetworkX has upgraded to 2.6.3 and add some new test cases. You can specify the `networkx==2.62` as  temporary solution.

Issue #1329, opened 2022-02-21
[proposal] Python API enhancement for compatibility of Giraph 


yecol:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

yecol:
These summaries your proposal and the discussion, do you have anything to add? @zhanglei1949

zhanglei1949:
> These summaries your proposal and the discussion, do you have anything to add? @zhanglei1949

I think your summary is quite clear, looking good to me.

siyuan0322:
I have some questions that may need to be clarified,

1. All apps except that return tensor context should return a graph? Not only the apps with vertex data context?
2. We could do this in almost pure python, we could extend the `run_app` with another select and `add_column` operation, then return the extended graph, (and project to a simple graph? I think this is not necessary as we have hidden the definition of projected-graph from user, and we have a implicit project operation before running app)
3. Apps may have more than one column as the result, such as hits. How to handle the case that
```python
g1 = hits(g)  # results have 2 columns (properties): auth, hub 
g2 = sssp(g1) # Cannot project because vertex have 2 properties.
```

PR #1352, opened 2022-03-07
Run app return graph

zhanglei1949:
## What do these changes do?

When a app runs on a simple graph and returns a `vertex data context`, we return a new graph.

## Related issue number

As a part of #1329 

Fixes


analytical_engine/core/grape_instance.h
```diff
@@ -89,7 +89,14 @@ class GrapeInstance : public Subscriber {
   bl::result<void> unloadApp(const rpc::GSParams& params);
 
   bl::result<std::string> query(const rpc::GSParams& params,
-                                const rpc::QueryArgs& query_args);
+                                const rpc::QueryArgs& query_args,
+                                const std::string& out_context_key,
+                                std::shared_ptr<IFragmentWrapper>& wrapper);
+
+  bl::result<rpc::graph::GraphDefPb> merge_ctx_to_new_graph(
```
sighingnow, line 96: Naming: camel case.

analytical_engine/core/fragment/arrow_projected_fragment.h
```diff
@@ -399,6 +402,52 @@ class ArrowProjectedFragment
...
+        projected_frag_meta.GetKeyValue<label_id_t>("projected_v_label");
+    VLOG(1) << "projected label id: " << projected_label_id;
+
+    // First get out arrow array
+    std::vector<std::pair<std::string, std::string>> selector_list;
+    selector_list.emplace_back("result", "r");
+    std::string selector_string = generate_selectors(selector_list);
+    auto selector = Selector::ParseSelectors(selector_string).value();
+    BOOST_LEAF_AUTO(arrow_arrays,
+                    ctx_wrapper->ToArrowArrays(comm_spec, selector));
```
sighingnow, line 423: The `ToArrowArray` will do a copy.

The copy can be avoid, but it's fine for current implementation for simplification. Please add a `TODO` or `FIXME` here.

sighingnow: We could allocate the context on vineyard directly, but its fine to keep the current codepath. Anyway add a TODO.

analytical_engine/core/grape_instance.cc
```diff
@@ -1097,8 +1121,24 @@ bl::result<std::shared_ptr<DispatchResult>> GrapeInstance::OnReceive(
...
-    BOOST_LEAF_AUTO(context_key, query(params, cmd->query_args));
-    r->set_data(context_key);
+    // For vertex context and arrowprojected fragment we return a new graph.
+    BOOST_LEAF_AUTO(graph_name, params.Get<std::string>(rpc::GRAPH_NAME));
+    BOOST_LEAF_AUTO(frag_wrapper,
+                    object_manager_.GetObject<IFragmentWrapper>(graph_name));
+    BOOST_LEAF_AUTO(return_graph, params.Get<bool>(rpc::RETURN_GRAPH));
+    std::string context_key = "ctx_" + generateId();
+    BOOST_LEAF_AUTO(context_result,
+                    query(params, cmd->query_args, context_key, frag_wrapper));
```
sighingnow, line 1131: Looks incorrect as the auto-assign `context_result` is never use.

sighingnow: I suggest don't touch the implementation of `query()` and use the returned `context_key` instead.

analytical_engine/core/grape_instance.h
zhanglei1949: Resolved

PR #1354, opened 2022-03-07
Limit the condition of ci actions to the alibaba/graphscope repo, as it may require self-hosted runners.

sighingnow:
## What do these changes do?

Limit the condition of ci actions to the `alibaba/graphscope` repo, as it may require self-hosted runners.


siyuan0322 (approved):
👍

PR #1369, opened 2022-03-14
Add data caching for networkx to improve the graph report performance

### LINKED ISSUE #386: [NetworkX-compatibility][Graph-CRUD] Optimization of networkx graph report methods
In networkx module, for simple and clear, we reuse the Views of NetworkX in graph report methods， but this take a side effect that like len(G.edges), len(G.nodes) would iterate all edges or nodes of graph. these meta actually can be report by g.number_of_nodes(), G.number_of_edges() at one time. 

Optimize the report methods by replace the iteration with more efficient op.

**Additional context**
part of #62

### LINKED ISSUE #1448: [NetworkX] Improve the report performance with cache strategy
The report methods like `nodes` of graphscope.nx fetch the node attribute and neighbors one after another and it's too slow.
The  performance of these methods can be improved by implement a cache strategy.

acezen:
## What do these changes do?
 This pull request add data caching for graphscope.nx graph to improve the graph report performance. It contains following changes:
- Use `Archive` to archive the report result;
- Add a cache to take care the iteration of graph and report data miss.
- Support tuple type node id with `MsgPack`

## Related issue number

Fixes #1448 
Fixes #386


python/graphscope/nx/classes/cache.py
```diff
@@ -0,0 +1,370 @@
...
+
+    def _get_node_id_cache(self, gid):
+        op = dag_utils.report_graph(
+            self._graph, types_pb2.NODE_ID_CACHE_BY_GID, gid=gid
+        )
+        archive = op.eval()
+        gid = archive.get_uint64()
+        node_size = archive.get_uint32()
+        fp = io.BytesIO(archive.get_bytes())
+        node_array = msgpack.load(fp, use_list=False)
```
sighingnow, line 327: msgpack could be `None`, as you have

```python
try:
    import msgpack
except ImportError:
    msgpack = None
import simdjson
```

sighingnow (changes requested):
I'm a bit unsure about if we need to introduce the msgpack dependency here, as we already have json as the serialization protocol, introducing a another one looks a bit redundant.

I think the issue here is json's inability to differ from tuple and list. But I cannot see how msgpack resolves that, as in `pack()` method in `msgpack_utils.h` I haven't found cases that related to tuple.

acezen:
> 

`msgpack` can parse array into `tuple` fast, not only list in json, that's what we need since list can't be key of hash map.  And `msgpack` unpack is faster than `simdjson` load.
https://github.com/msgpack/msgpack-python#how-to-use

python/graphscope/nx/classes/cache.py
acezen: thanks, I'll fix it

sighingnow (approved):
> in the future, we can replace json with msgpack protocol somehow.
https://github.com/msgpack/msgpack-python#how-to-use

As long as we have some (micro) benchmarks that could show msgpack is faster than json (both in the C++ side and Python side).

lidongze0629 (approved):

sighingnow (approved):

sighingnow (approved):

PR #1389, opened 2022-03-22
Upgrade vineyard to v0.3.21.

sighingnow:
## What do these changes do?

The latest vineyardd is not compatible with the older client, as the support for reference counting.


siyuan0322 (approved):

sighingnow:
The networkx test under distributed setting are failed and beyond the scope of this "upgrading" PR. I have disabled the test and will open another PR to track that.

PR #1457, opened 2022-04-14
Correctly handle the empty DAG

siyuan0322:
Submitting empty DAG should not trigger some unhandled case such as no such attribute or index out of range.

### DIFF
--- a/coordinator/gscoordinator/dag_manager.py
+++ b/coordinator/gscoordinator/dag_manager.py
@@ -88,26 +88,27 @@ class DAGManager(object):
                 req_head = req
             else:
                 req_bodies.append(req)
-        # split dag
-        dag = op_def_pb2.DagDef()
-        dag_for = GSEngine.analytical_engine
-        dag_bodies = []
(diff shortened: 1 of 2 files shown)


coordinator/gscoordinator/dag_manager.py
```diff
@@ -88,26 +88,27 @@ def __init__(self, request_iterator: Sequence[message_pb2.RunStepRequest]):
...
+                    dag_for = self.get_op_exec_engine(op)
+                    dag_bodies = []
+                # select op
+                dag.op.extend([copy.deepcopy(op)])
+                for req_body in req_bodies:
+                    # select chunks belong to this op
+                    if req_body.body.op_key == op.key:
+                        dag_bodies.append(req_body)
+            if dag.op:
+                self._dag_queue.put((dag_for, dag, dag_bodies))
```
sighingnow, line 111: I suggest adding

```python
if req_head is None:
    return
```

instead to minimize the code changeset.

sighingnow (approved):
LGTM.

PR #1460, opened 2022-04-15
Correctly handle the lifecycle of graph

### LINKED ISSUE #1394: Double check the lifecycle of context.
Ensure the lifecycle of context is correct to prevent memory-leak or invalid access.

- [x] The context must hold the graph even if the graph got unloaded.
- [x] The context memory can be freed by hand or when the context gets deleted.
- [x] If a graph is only held by a context, the unloading of the context should also unload the graph.

siyuan0322:
- Use del rather than unload
- Projected graph now hold a reference to the base graph

Fixes #1394

### DIFF
--- a/python/graphscope/framework/graph.py
+++ b/python/graphscope/framework/graph.py
@@ -77,10 +77,6 @@ class GraphInterface(metaclass=ABCMeta):
     ):
         raise NotImplementedError
 
-    @abstractmethod
-    def unload(self):
-        raise NotImplementedError
-
     def to_numpy(self, selector, vertex_range=None):
         raise NotImplementedError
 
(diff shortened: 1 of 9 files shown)


lidongze0629 (approved):

python/graphscope/framework/graph.py
```diff
@@ -77,10 +77,6 @@ def add_edges(
     ):
         raise NotImplementedError
 
-    @abstractmethod
-    def unload(self):
-        raise NotImplementedError
```
sighingnow, line 82: Please keep the `unload()` API, and raising a warning to tell users about the deprecation, like

```
warnings.warn("The Graph.unload() method has been deprecated, please using the `del` operator instead, e.g., `del graph`", DeprecationWarning)
```

sighingnow (changes requested):
And, I suggest using `del` instead of private `_unload()` in affected test cases.

siyuan0322:
> And, I suggest using `del` instead of private `_unload()` in affected test cases.

These remaining `_unload()` is for lazy... you must have a return value for session to run...

sighingnow:
> > And, I suggest using `del` instead of private `_unload()` in affected test cases.
> 
> These remaining `_unload()` is for lazy... you must have a return value for session to run...

We could `sess.eval(g._unload())` in `_del` in both cases.

sighingnow (approved):

PR #1480, opened 2022-04-24
Add another write path that uses the C++ IO adaptor

### LINKED ISSUE #1468: Support extending external drivers for write.
The reading graph process has been extended to use customer drivers by registering environment variable, the write process should also support similar function.

siyuan0322:
- Add another write path that uses the C++ IO adaptor
- Reduce some code duplication in `grape_instance.cc`
- Move ArrowFlattenProject related stuff out of the NETWORKX guard.

Fixes #1468

### DIFF
--- a/analytical_engine/core/grape_instance.cc
+++ b/analytical_engine/core/grape_instance.cc
@@ -353,21 +353,11 @@ bl::result<void> GrapeInstance::modifyEdges(const rpc::GSParams& params) {
 
 bl::result<std::shared_ptr<grape::InArchive>> GrapeInstance::contextToNumpy(
     const rpc::GSParams& params) {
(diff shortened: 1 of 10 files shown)


analytical_engine/core/grape_instance.cc
```diff
@@ -728,11 +688,114 @@ bl::result<std::string> GrapeInstance::contextToVineyardDataFrame(
   return toJson({{"object_id", s_id}});
 }
 
+bl::result<void> GrapeInstance::outputContext(const rpc::GSParams& params) {
```
acezen, line 691: Overall LGTM, you can extract method to fix the CodeFactor issue and make the code more readable.

sighingnow (approved):

PR #1511, opened 2022-05-06
Make client and coordinator setup with develop in MakeFile

acezen:
## Related issue number

Fixes


sighingnow:
I think `setup` client/coordiantor is not a necessity for running local graphscope code, as we already have much effort for inplace running, e.g., client can find coordinator with relative path `../coordinator`.

Besides, `python3 setup.py develop` is not a good idea as there's some issues during dependency installation process (e.g., https://github.com/pypa/setuptools/issues/2864), and, won't be fixed by setuptools. Instead, `pip3 install -e .` would be prefer (`-e` means `--editable`).

lidongze0629 (approved):

sighingnow (approved):

PR #1515, opened 2022-05-06
add ci tests for ir on vineyard

shirly121:
## Related issue number

Fixes


sighingnow (approved):

lidongze0629 (approved):

.github/workflows/gaia.yml
```diff
@@ -65,6 +65,11 @@ jobs:
         cd ${GITHUB_WORKSPACE}/research/query_service/ir/compiler
         make build
 
+    - name: Build Ir on Vineyard Store
+      run: |
+        source ${HOME}/.bashrc
```
yecol, line 70: Why do we need to source this bashrc? It is user's customed envs. 
If you have envs, should put them together into the [.graphscope_env](https://github.com/alibaba/GraphScope/blob/main/scripts/install_deps.sh#L20)

yecol (changes requested):
See comments.

.github/workflows/gaia.yml
```diff
@@ -86,3 +91,9 @@ jobs:
         source ${HOME}/.bashrc
         cd ${GITHUB_WORKSPACE}/interactive_engine/ir-adaptor && ./ir_groot_ci.sh
 
+    - name: Ir Integration Test on Vineyard Store
+      run: |
+        source ${HOME}/.bashrc
```
yecol, line 96: ditto

.github/workflows/gaia.yml
sighingnow: I'm not sure why, but we do have many `source ~/.bashrc` lines across all CI yamls. It might be used for leveraging local build cache or something like that, I guess.

siyuan0322 (approved):

.github/workflows/gaia.yml
shirly121: done

yecol (approved):

Issue #1499, opened 2022-04-29
DevInfra enhancement


yecol:
as titled. details TBF.

PR #1545, opened 2022-05-11
compile pass

### LINKED ISSUE #1544: gie sdk refactor
as titled

tianliplus:
## Related issue number

Fixes #1544


interactive_engine/compiler/src/main/java/com/alibaba/maxgraph/compiler/tree/BranchTreeNode.java
```diff
@@ -1,16 +1,14 @@
...
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
+ * <p>Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing permissions and
  * limitations under the License.
  */
 package com.alibaba.maxgraph.compiler.tree;
```
longbinlai, line 14: I think `maxgraph` better be replace with graphscope.gie, namely com.alibaba.graphscope.gie...

longbinlai (commented):
Is it a better idea to include an sdk-common-v2 that include all the new codes without touching the legacy ones? While we complete the replacement of maxgraph with gaia, we then remove old sdk-common @yecol @tianliplus

interactive_engine/compiler/src/main/java/com/alibaba/maxgraph/compiler/tree/BranchTreeNode.java
yecol: Yes, I cannot agree more. 

Actually, I also found these in shirly121's PR and recorded in #1499 (the last line).

yecol:
sdk-common-v2 is a good idea! 
I agree!

tianliplus:
There are a lot of "com.aliababa.maxgraph" in our project. I think we should use a dedicated issue to track the refactoring work. see #1564

longbinlai (approved):
Another issue is created for renaming the package of sdk-common.

PR #1546, opened 2022-05-11
Clean up include headers of analytical engine

### LINKED ISSUE #1541: Clean up includes to reduce the building time for analytical engine
as titled. 
use tools like https://github.com/include-what-you-use/include-what-you-use

acezen:
## What do these changes do?
- Clean up headers of GAE with iwyu
- Remove leaf header from `core/config.h`, put to where leaf been used
- analytical engine building time from 82.89s to 72.26s, size from 1889376 bytes to 1815784 bytes (on my own mac with `make` and remove  gs_proto building time)

## Related issue number

Fixes #1541


sighingnow (commented):
Why `bl::` -> `boost::leaf`?

acezen:
> Why `bl::` -> `boost::leaf`?

The original 'bl' namespace is define in `core/config.h`, include the `boost/leaf.hpp`, and it would make some code include `core/config.h` but not use `boost/leaf` also include the leaf headers. That breaks the include what you use rule.

yecol:
> Why `bl::` -> `boost::leaf`?

I have the same question.

yecol:
> > Why `bl::` -> `boost::leaf`?
> 
> The original `bl` namespace is define in `core/config.h`, include the `boost/leaf.hpp`, and it would make some code include `core/config.h` but not use `boost/leaf` also include the leaf headers. That breaks the include what you use rule. So I remove the leaf headers from config.h, put to where leaf been used and recover `bl` to `boost::leaf`.

One of the solutions to this is defining the `bl` alias at the beginning of each used file.
Too long nested namespace has harmful impact on the code readability.

yecol:
Could you also report the effectiveness of this PR to speed up the building stage?
e.g., from XXs -> YYs.

yecol (approved):

PR #1552, opened 2022-05-13
Implement voterank app 

### LINKED ISSUE #1526: A pagerank-variant algorithm: voterank with `PIE` model
Implement a pagerank-variant algorithm: voterank with `PIE` model

VoteRank  computes a ranking of the nodes in a graph G based on a voting scheme. With VoteRank, all nodes vote for each of its in-neighbours and the node with the highest votes is elected iteratively. The voting ability of out-neighbors of elected nodes is decreased in subsequent turns.

Paper: https://www.nature.com/articles/srep27823

Reference implementation: https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.centrality.voterank.html?highlight=voterank

liulx20:
## What do these changes do?

Implement voterank app.

- [x] implement C++ PIE app with analytical engineAPI;
- [x] check the correctness with NetworkX;
- [ ] check the scalability and performance in large graph;
- [ ] add to python built-in algorithms and add related test case

## Related issue number
Fixes #1526


analytical_engine/apps/voterank/voterank.h
```diff
@@ -0,0 +1,224 @@
...
+*/
+
+#ifndef EXAMPLES_ANALYTICAL_APPS_VOTERANK_VOTERANK_H_
+#define EXAMPLES_ANALYTICAL_APPS_VOTERANK_VOTERANK_H_
+
+#include <grape/grape.h>
+
+#include "voterank/voterank_context.h"
+
+namespace grape {
```
acezen, line 23: analytical apps in GraphScope should use namespace gs, not grape

acezen (changes requested):
Good job for the implementation. There are still something need to revise:
- Format the code to pass the CI check;
- Add test case for voterank app, you can add a p2p_31 test to [test_ctx_builtin.py](https://github.com/alibaba/GraphScope/blob/main/python/graphscope/nx/tests/test_ctx_builtin.py) and migrate the [networkx test](https://github.com/networkx/networkx/blob/main/networkx/algorithms/centrality/tests/test_voterank.py) to GraphScope;

analytical_engine/apps/voterank/voterank.h
```diff
@@ -0,0 +1,224 @@
...
+              }
+          };
+    // compute new scores
+    std::vector<std::pair<double,oid_t>> max_scores(thread_num(),{0,{}});
+    ForEach(inner_vertices, [&ctx,compare, &max_scores,&frag](int tid, vertex_t u) {
+        if (ctx.rank[u] == 0) {
+          double cur = ctx.scores[u];
+          auto es = frag.GetOutgoingOuterVertexAdjList(u);
+          for (auto& e : es) {
+            cur += ctx.weight[e.get_neighbor()];
```
acezen, line 161: would that better if  we compute the weight sum one time and keep in a context member ?  It could avoid to recompute the weight sum of u

acezen (approved):
LGTM

PR #1597, opened 2022-05-23
Optimize clustering by ignoring computation of vertices that degree less than 2

### LINKED ISSUE #1586: [Proposal]Performance optimization in clustering coefficient
@yecol 

I got a very bad performance when I use the clustering coefficient api over a very large dataset.  

Then I realized that there are lots of nodes in the graph with only 1 neighbor which theoretically means that their clustering coefficient is zero.

So I think if there is any possibility to reduce the computation cost over the nodes with degree of 1, in order to improve the clustering coefficient performance.

acezen:
## What do these changes do?
Since vertices which degree < 2 its clustering coefficient is 0 , we can optimize clustering by ignoring the computation of this kind of vertices.

## Related issue number

Fixes #1586


acezen:
cc/ @doudoubobo please take a review of the pull request

sighingnow (approved):

yecol (approved):

PR #1608, opened 2022-05-24
fix the join bug

### LINKED ISSUE #1613: [BUG] Join operator sometimes may miss the result.
**Describe the bug**
Join operator sometimes may miss the result.
**To Reproduce**

1. Generator two streams, each with 1000 elements whose key=1.
2. Join two streams and count the number of matched pairs.

Details are illustrated in the following codes.

```
#[test]
fn join_test_with_same_key() {
    let mut conf = JobConf::new("inner_join");
    conf.set_workers(1);
    let mut result = pegasus::run(conf, || {
        let id = pegasus::get_current_worker().index;
        move |input, output| {
            let (src1, src2) = input.input_from(0..1000 as i32)?.copied()?;
            src1.key_by(|x| Ok((1 as i32, x)))?
                .partition_by_key()
                .inner_join(src2.key_by(|x| Ok((1 as i32, x)))?.partition_by_key())?
                .map(|(d1, d2)| Ok(((d1.key, d1.value), (d2.key, d2.value))))?
                .collect::<Vec<((i32, i32), (i32, i32))>>()?
                .sink_into(output)
        }
    })
        .expect("run job failure;");

    let mut result = result.next().unwrap().unwrap();
    assert_eq!(result.len(), 1000*1000);
}
```

**Expected behavior**
Outputs 1,000,000 pairs, but actually misses parts of results.

wzbxpy:
## What do these changes do?

The join operator in Pegasus sometimes misses some results. 
The reason is that applying session.give() to output the join results one by one will throw away the data that overflow the cache of the stream.
We replace it but session.give_iterator() and fix the bug.

## Related issue number

Fixes #1613

### DIFF
--- a/research/engine/pegasus/pegasus/src/operator/concise/keyed/join.rs
+++ b/research/engine/pegasus/pegasus/src/operator/concise/keyed/join.rs
@@ -91,8 +91,8 @@ impl<L: Data + HasKey, R: Data + HasKey> Helper<L, R> {
 fn insert_and_query<'a, L: Data + HasKey, R: Data + HasKey<Target = L::Target>>(
     map1: &mut JoinMap<L>, map2: &'a mut JoinMap<R>, data: &L, need_insert: bool,
 ) -> Option<&'a Vec<R>>
-where
-    L::Target: Clone + Send,
+    where
+        L::Target: Clone + Send,
 {
     let k = data.get_key();
     let entry1 = map1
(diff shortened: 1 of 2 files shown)


longbinlai (approved):
LGTM

wzbxpy:
> hi @wzbxpy, thanks for your first commit to this repo! Please add a brief intro in the summary (even a sentence) and link this PR to the related issue. :)

Added.

PR #1663, opened 2022-06-06
Fix the nightly ci, including build wheel on macOS/linux and networkx forward algorithm

acezen:
## Related issue number

Fixes #1644

PR #1681, opened 2022-06-06
Fixes the nx.compat for name conflicts

sighingnow:
## What do these changes do?

There are name conflicts in networkx, e.g.,

- `networkx.algorithms.approximation.connectivity` vs. `networkx.algorithms.connectivity`

## Related issue number

N/A


sighingnow:
Close as in favour of #1663.

PR #1709, opened 2022-06-13
[GAIA-IR] Reorg for GAIA-IR's Runtime

### LINKED ISSUE #1708: Reorg for GAIA-IR's Runtime
Currently, the `GraphProxy` Trait for querying storage is defined in the `runtime` crate. A `graph_proxy` crate is further defined to implement `GraphProxy` for different storages, depending on both `runtime` and `storage`, in consideration of a loose coupling of the Runtime and Storage components. 
However, when introducing the Trait for writing a graph to storage after querying in runtime, the runtime crate has to depend on the storage. To make the dependency more clear, we reorg the runtime related crates in the following way: 
1. We define `GraphProxy` and related APIs like `Vertex`, `Edge`, `Property` etc. in `graph_proxy` crate, together with the implementations for different storages. Futhermore, we will define the `WriteGraphProxy` trait for writing a graph to storage in this crate as well.
2. `runtime` crate will depend on `graph_proxy` crate, to query from storages and write into storages. And runtime will provide a `IRJobAssembly` implementation, who will translate IR ops into udf (executed in engine), including the ops that will query the storage.
3. We define a `InitializeJobAssembly` trait in the `runtime_integration` crate, to initialize a `IRJobAssembly` when starting engine with the specified storage.

BingqingLyu:
## Related issue number

Fixes #1708


longbinlai (commented):
1) `graph_proxy/src/graph_proxy.rs` -> should be in `graph_proxy/src/mod.rs`, thus we donot have the package name like `graph_proxy::api::graph_proxy`
2) the module name of `api` -> better be `apis`
3) Considering place `exp_store` and `gs_store` under a module named adaptors under graph_proxy. 
4) Why gs_store, can we name it `groot_store` ??

BingqingLyu:
> 1. `graph_proxy/src/graph_proxy.rs` -> should be in `graph_proxy/src/mod.rs`, thus we donot have the package name like `graph_proxy::api::graph_proxy`
> 2. the module name of `api` -> better be `apis`
> 3. Considering place `exp_store` and `gs_store` under a module named adaptors under graph_proxy.
> 4. Why gs_store, can we name it `groot_store` ??

1/2/3: Done. 
4: it is actually short for graphscope_store for now. After refining the read path, i.e., directly adapt for vineyard, it will become groot_store then.

PR #1727, opened 2022-06-16
Supports launch vineyardd using on multiple hosts using MPI

sighingnow:
## What do these changes do?

The `addrs` argument still be useful in some cases, however in such cases vineyard need to be launched properly on multiple hosts as well, using MPI as the launch should be ok.

## Related issue number

N/A


lidongze0629 (approved):

zhanglei1949 (approved):
LGTM.

PR #1743, opened 2022-06-21
[IR Compiler] fix error when creating json file

### LINKED ISSUE #1729: [BUG] Panicked at 'Create json file': 'plan.json' error: Read-only file system
**Describe the bug**

The path of 'plan.json' is relative to the installation directory, which may cause permission problems. It's better to change the path to `~/.graphscope` or `/tmp`, and use a unique ID to distinguish the different queries.

image

shirly121:
… to /tmp instead of current to have access to

## What do these changes do?
1. write json file to /tmp instead of the current directory to solve the problem of file privilege
2. attach job id to each json file name to avoid conflicts in concurrent scenarios

## Related issue number

Fixes #1729

### DIFF
--- a/research/query_service/ir/compiler/src/main/java/com/alibaba/graphscope/common/IrPlan.java
+++ b/research/query_service/ir/compiler/src/main/java/com/alibaba/graphscope/common/IrPlan.java
@@ -46,8 +46,8 @@ import java.util.function.Function;
 // represent ir plan as a chain of operators
 public class IrPlan implements Closeable {
     private static IrCoreLibrary irCoreLib = IrCoreLibrary.INSTANCE;
-    // write to '/tmp' which is accessible from common users
-    private static String JSON_PLAN_DIR = "/tmp";
+    // get tmp directory from system properties, which is accessible from common users
+    private static String JSON_PLAN_DIR = System.getProperty("java.io.tmpdir");
     private Pointer ptrPlan;
     private IrMeta meta;
     // to identify a unique json file which contains the logic plan from ir_core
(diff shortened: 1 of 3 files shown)


lidongze0629 (approved):

research/query_service/ir/compiler/src/main/java/com/alibaba/graphscope/common/IrPlan.java
```diff
@@ -46,9 +46,12 @@
 // represent ir plan as a chain of operators
 public class IrPlan implements Closeable {
     private static IrCoreLibrary irCoreLib = IrCoreLibrary.INSTANCE;
-    private static String PLAN_JSON_FILE = "plan.json";
+    // write to '/tmp' which is accessible from common users
+    private static String JSON_PLAN_DIR = "/tmp";
```
longbinlai, line 50: Can we obtain temporary fold from some environment variables. For example, Mac's temporary fold is not at "/tmp"

lidongze0629: I think `/tmp` is ok in macOS, which is a link to `/private/tmp`, Or maybe you can get the tmp directory by `System.getProperty('java.io.tmpdir')`

> https://kodejava.org/how-do-i-get-operating-system-temporary-directory-folder/

```java
package org.kodejava.lang;

public class TempDirExample {
    public static void main(String[] args) {
        // This is the property name for accessing OS temporary directory
        // or folder.
        String property = "java.io.tmpdir";

        // Get the temporary directory and print it.
        String tempDir = System.getProperty(property);
        System.out.println("OS temporary directory is " + tempDir);
    }
}
```

shirly121: done

PR #1788, opened 2022-07-01
impl PartialOrd and some methods for Property

waruto210:
## What do these changes do?

The derived `PartialOrd` of enum is not suitable, impl PartialOrd and some methods for Property, then format it.

## Related issue number

None

### DIFF
--- a/interactive_engine/executor/store/src/api/property.rs
+++ b/interactive_engine/executor/store/src/api/property.rs
@@ -1,12 +1,12 @@
 //
 //! Copyright 2020 Alibaba Group Holding Limited.
-//! 
+//!
 //! Licensed under the Apache License, Version 2.0 (the "License");
 //! you may not use this file except in compliance with the License.
 //! You may obtain a copy of the License at
-//! 
+//!
 //! http://www.apache.org/licenses/LICENSE-2.0
-//! 
+//!
 //! Unless required by applicable law or agreed to in writing, software
 //! distributed under the License is distributed on an "AS IS" BASIS,
 //! WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


interactive_engine/executor/store/src/api/property.rs
```diff
@@ -44,6 +44,51 @@ pub enum Property {
...
+            // if both are integers, cast to long
+            // else cast to double
+            (Property::Short(_), _)
+            | (Property::Int(_), _)
+            | (Property::Long(_), _)
+            | (_, Property::Short(_))
+            | (_, Property::Int(_))
+            | (_, Property::Long(_)) => {
+                if self.is_float_type() || other.is_float_type() {
+                    let left = self.get_double().expect("Convert error");
```
BingqingLyu, line 62: It's better not to use `expect()` here, which will panic once Property is neither integers nor floats. Same as the follows.

interactive_engine/executor/store/src/api/property.rs
```diff
@@ -44,6 +44,51 @@ pub enum Property {
...
+            (Property::Float(left), Property::Float(right)) => left.partial_cmp(right),
+            (Property::Double(_), _) | (_, Property::Double(_)) => {
+                let left = self.get_double().expect("Convert error");
+                let right = other.get_double().expect("Convert error");
+                left.partial_cmp(&right)
+            }
+            (Property::Bytes(left), Property::Bytes(right)) => left.partial_cmp(right),
+            (Property::String(left), Property::String(right)) => left.partial_cmp(right),
+            (Property::Date(left), Property::Date(right)) => left.partial_cmp(right),
+            (Property::ListInt(left), Property::ListInt(right)) => left.partial_cmp(right),
```
BingqingLyu, line 80: As above, we compare integers/floats by converting into `long`/`double`. How about compare between  e.g., `Property::ListInt` and `Property::ListLong`?

BingqingLyu (commented):
As above.

interactive_engine/executor/store/src/api/property.rs
waruto210: introduce a macro to handle it

interactive_engine/executor/store/src/api/property.rs
waruto210: now cast number lists to cmp

waruto210:
> As follows.

please review again

BingqingLyu (approved):
LGTM.

Issue #1772, opened 2022-06-28
[GIE/Gaia-IR] Compile IR with v6d dependency errors


longbinlai:
After merging the latest version of Gaia, the v6d dependency errors prompt on Gaia/IR, which cannot be compiled without installing v6d as a dependency, even when v6d is not needed at all, such as in the unit tests by `cargo test`. Such dependency errors should be avoided.  

To reproduce, running  `cargo test`  under  research/query_service/ir shows: 
 ```
--- stderr
  CMake Error at CMakeLists.txt:53 (find_package):
    By not providing "Findvineyard.cmake" in CMAKE_MODULE_PATH this project has
    asked CMake to find a package configuration file provided by "vineyard",
    but CMake did not find one.

    Could not find a package configuration file provided by "vineyard"
    (requested version 0.4.1) with any of the following names:

      vineyardConfig.cmake
      vineyard-config.cmake

    Add the installation prefix of "vineyard" to CMAKE_PREFIX_PATH or set
    "vineyard_DIR" to a directory containing one of the above files.  If
    "vineyard" provides a separate development package or SDK, be sure it has
    been installed.
```

sighingnow:
You need install vineyard first.

BingqingLyu:
- [ ] Should make all codes related to v6d_ffi as features, including those in GAIA-Runtime.
- [ ] It seems that `dyld error` occur on macos in some cases.

sighingnow:
> It seems that dyld error occur on macos in some cases.

It was caused by the difference in the behaviour of cargo's LTO setting. Has discussed with @MeloYang05 for a workaround.

BingqingLyu:
> > It seems that dyld error occur on macos in some cases.
> 
> It was caused by the difference in the behaviour of cargo's LTO setting. Has discussed with @MeloYang05 for a workaround

PR #1795, opened 2022-07-05
[IR Runtime] Make v6d related codes as features

BingqingLyu:
## What do these changes do?

As titled.

## Related issue number

#1772


longbinlai (commented):
Do you have some tests that  only passed with v6d? Otherwise should add some

BingqingLyu:
> Do you have some tests that only passed with v6d? Otherwise should add some

It seems that we can only add integration tests on query based on v6d for now, since function of `create_v6d_builder` is only available in gs coordinator, while we can only `get_graph_builder` in `v6d_ffi`.

longbinlai (approved):

Issue #1798, opened 2022-07-05
Pegasus cluster config files should support hostname instead of only IP


MeloYang05:
**Is your feature request related to a problem? Please describe.**
In GraphScope's current version, starting a pegasus cluster needs to know all servers' IPs. It is not friendly in a cloud-native environement. For example, if we want to start a pegasus cluster with kubernates, we have to use kubernates' dns server to know each server's IP and edit the config toml file in each container before starting the cluster. It is not only inconvenient but also ugly. Therefore, I strongly recommended to support using hostname in pegasus cluster's config file.

**Describe the solution you'd like**
Support using hostname in pegasus cluster's config file. E.g., we can write localhost in the toml file instead of 127.0.0.1.
***current requirement:***
```
[[network.servers]]
ip = '127.0.0.1'
port = 11234
```
***expected one:***
```
[[network.servers]]
ip = 'localhost'
port = 11234

```

MeloYang05:
The issue is solved

PR #1808, opened 2022-07-06
Make Pegasus support reading config files with hostname instead of only IP

MeloYang05:
## What do these changes do?
Make Pegasus support reading config files with hostname instead of only IP

## Related issue number
https://github.com/alibaba/GraphScope/issues/1798

Fixes


longbinlai:
The solution is handy, better let pegasus network connection works with hostname rather than ipaddr.

bmmcq (approved):

Issue #1802, opened 2022-07-06
Support Tinkerpop gremlin test cases in GAIA IR.


siyuan0322:
**We plan to support these test cases in this scrum (0627 - 0715)**
- [x] aggregation, min / max / sum / countDistinct / toList(fold) (to be fixed in `ir_compiler_aggFn`)
- [x] mean (avg in ir)
- [x] hasNot
- [x] dedup by variables
- [x] dedup by subtask
- [x] selecting a none tag should return empty result instead of throwing exception (to be fixed in `select_none_tag`)
- [x] cases with unexpected results (`range(1,2)` with a lower bound greater than 0 is unsupported in runtime)

**We plan to support these test cases in the near future**
- [ ] and/or operators
- [ ] count(local)/order(local)
- [ ] repeat
- [ ] path, simplePath, cyclicPath

**Support for these test cases are not planned**
1. choose
2. parameters of V() and hasId() are constrained to integer or list of integers in ir grammar, V("1") or hasId("id") will not be supported under this consideration

siyuan0322:
Remove the TODO in [ci.yml](https://github.com/alibaba/GraphScope/blob/a675598e54c74ce9699de84718f09cd334bf6956/.github/workflows/ci.yml#L620)

longbinlai:
Remaining cases such as path/repeat etc may be reopened in a new issue in the next-stage scrum.

PR #1823, opened 2022-07-11
fix groot ci issue  after replacing with gaia

### LINKED ISSUE #1804: Pass the groot-server tests with gaia engine executor.
The `mvn test` in groot-server was using maxgraph executor in the past.
After switched to gaia executor, some tests will raise NullPointerException.

shirly121:
## Related issue number
as titled.

Fixes #1804
Link #1802


siyuan0322 (approved):
👍

Issue #1864, opened 2022-07-21
[BUG] Improve the compilation time and memory consumption when building GAE


sighingnow:
**Describe the bug**

Seems github has (may be temporarily) adjust the memory limit of runners on Github Actions, causing failure on our main branch: https://github.com/alibaba/GraphScope/runs/7441676827?check_suite_focus=true

The building time and memory usage are indeed a longstanding problem of GAE. I would take a look and try simplifying the header/module dependencies.

**Additional context**

sighingnow:
Closing as resolved.

PR #1868, opened 2022-07-22
Decrease the make job numbers when building gae.

siyuan0322:
This is a temporary solution to make the CI pass.
Change this back when #1864 is merged.


siyuan0322:
The build duration increased from 47 min to 65 min.

PR #1883, opened 2022-07-25
[GIE] Reorganize executor package by incorporating GAIA into GIE

BingqingLyu:
## What do these changes do?

Reorganize executor package by moving GAIA from the `research` directory into `interactive_engine`, including reorganizing packages in a more reasonable way, removing unused packages, etc.

## Related issue number

#1677

PR #1902, opened 2022-07-28
Format code in research

waruto210:
## What do these changes do?

make format of dyn_type and graph_store the same with query_service
## Related issue number

None


BingqingLyu:
Have formatted in #1883 :)

PR #1908, opened 2022-07-31
impl partial_eq for Property and fix gitignore

waruto210:
## What do these changes do?
impl partial_eq for Property and add deps for build tests and benches, then fix gitignore

## Related issue number

None


BingqingLyu (approved):
LGTM.

PR #1922, opened 2022-08-02
Fixed graph schema drop and init error when edge_kinds's src or dst gone

iamdrq:
## What do these changes do?

Fixed graph schema drop, origin labels return string but expect get object 
Fixed init error when edge_kinds's src or dst gone, because can't get label from id_to_label by id


siyuan0322 (approved):

siyuan0322:
@iamdrq  Merged, Thanks!

PR #1925, opened 2022-08-03
[GAIA-IR] create ffi object from java side to avoid invoking ffi interface which can lead memory leak problems

shirly121:
## What do these changes do?
as titled

## Related issue number

Fixes


longbinlai (approved):
LGTM

PR #1926, opened 2022-08-03
Fixed ingestor high CPU usage when empty buffer

iamdrq:
## What do these changes do?

Fixed ingestor high CPU usage when empty buffer, because no wait loop happened


siyuan0322 (approved):
LGTM

siyuan0322:
Hi @iamdrq, could you rebase the PR and then we can get it landed.

Issue #1885, opened 2022-07-26
[BUG] Pegasus server fail to restart after failover


lnfjpt:

longbinlai:
Reopen due to: https://github.com/alibaba/GraphScope/pull/1969

PR #1950, opened 2022-08-08
fix: add ipc sender after former one removed

lnfjpt:
## Related issue number

Fixes
https://github.com/alibaba/GraphScope/issues/1885


longbinlai (approved):
LGTM

PR #1969, opened 2022-08-15
fix network connection reconnect abnormal

bmmcq:
see #1885 and #1950


siyuan0322 (approved):
Tested on cluster with 3 nodes, killing 1, 2, 3 workers multiple times and these could reconnected with each other as expected.

PR #2005, opened 2022-09-05
[GIE-IR] support SimplePath and AllV in path expand operator

shirly121:
## What do these changes do?
as titled

## Related issue number

Fixes


interactive_engine/compiler/src/test/java/com/alibaba/graphscope/gremlin/PathExpandStepTest.java
```diff
@@ -60,4 +58,20 @@ public void g_V_path_expand_label_test() {
         FfiNameOrId.ByValue label = op.getParams().get().getTables().get(0);
         Assert.assertEquals("knows", label.name);
     }
+
+    // g.V().out("1..2").with("PathOpt", "Simple").with("ResultOpt", "AllV")
+    @Test
+    public void g_V_path_expand_with_test() {
+        Traversal traversal =
+                g.V().out(__.range(1, 2)).with("PathOpt", "Simple").with("ResultOpt", "AllV");
```
longbinlai, line 66: Can we use enum instead of string to configure these "PathOpt", "Simple", "ResultOpt", "AllV"?

interactive_engine/executor/ir/graph_proxy/src/apis/graph/element/path.rs
```diff
@@ -44,48 +47,84 @@ impl From<Edge> for VertexOrEdge {
     }
 }
 
-#[derive(Clone, Debug, Hash)]
+#[derive(Clone, Debug)]
 pub enum GraphPath {
-    WHOLE((Vec<VertexOrEdge>, usize)),
-    END((VertexOrEdge, usize)),
+    AllV(Vec<VertexOrEdge>),
+    SimpleAllV((Vec<VertexOrEdge>, HashSet<ID>)),
```
longbinlai, line 53: Do not think it is a good idea to store `ID` in a hashset, which is very costly and memory-inefficient. Considering that the path often contains less than 4 items, simple iterate the vector to `dedup()` is not very unacceptable. Same for `SimpleEndV`

longbinlai (approved):
LGTM

PR #2009, opened 2022-09-06
Ensure session close work when triggered not in main thread

dzhiwei:
## What do these changes do?

Signal is not working when the running program is not in python main thread. ensure session close can work in above case.

## Related issue number

Fixes


sighingnow (approved):
LGTM.

yecol:
hi @dzhiwei could you sign the CLA so that we can merge this PR?

PR #2012, opened 2022-09-07
Add connected() in session.py to figure out session status

dzhiwei:
GAE add connected() in session.py to indicate if the session is still valid and available for the graph computing work. In some case analytical engine cannot be connected, with this new function, client side can check the session status and process this special case(try to restart runtime or close session etc).

## What do these changes do?

GAE add connected() in session.py to indicate if the session is still valid and available for the graph computing work.
In some case analytical engine cannot be connected, with this new function, client side can check the session status and process this special case(try to restart runtime or close session etc).

## Related issue number

Fixes

### DIFF
--- a/python/graphscope/client/session.py
+++ b/python/graphscope/client/session.py
@@ -802,6 +802,14 @@ class Session(object):
                 else:
                     self._disconnected = False
             time.sleep(self._heartbeat_interval_seconds)
+            
+    def connected(self) -> bool:
+        """Check if the session is still connected and available.
+
+        Returns: True or False
+
+        """
+        return not self._disconnected
 
     def close(self):
         """Closes this session.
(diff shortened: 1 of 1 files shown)


python/graphscope/client/session.py
```diff
@@ -802,6 +802,14 @@ def _send_heartbeat(self):
                 else:
                     self._disconnected = False
             time.sleep(self._heartbeat_interval_seconds)
+
```
sighingnow, line 805:
```suggestion

```

python/graphscope/client/session.py
```diff
@@ -802,6 +802,13 @@ def _send_heartbeat(self):
                 else:
                     self._disconnected = False
             time.sleep(self._heartbeat_interval_seconds)
+    def connected(self) -> bool:
```
sighingnow, line 805:
```suggestion

    def connected(self) -> bool:
```

sighingnow (approved):

Issue #2013, opened 2022-09-07
[BUG] bug in label encoding of `exp_store`


BingqingLyu:
**Describe the bug**

Run the following query with `exp_store` on  LDBC dataset outputs empty results, which is not as expected:
`g.V().has("COUNTRY", "name", "India").in("ISPARTOF").hasLabel("CITY")`

This is caused by the hierarchical label structure (two-layers) in `exp_store`. For example, a "PERSON" would be encoded as `[PERSON, INVALID_LABEL_ID]`, while a "CITY" whould be encoded as `[PLACE, CITY]`. However, we only take the firstly layer of label for now. e.g., when compared with `CITY`, we take label of `PLACE`, which leads to the wrong answer.

PR #2014, opened 2022-09-08
[Bug Fix] fix bug in label encoding of `exp_store`

BingqingLyu:
## What do these changes do?

As titled.

## Related issue number

#2013


BingqingLyu:
This would cause unexpected results for bi-11. Here, we fix label encoding in GraphProxy. In the future, it would be fixed in ExpStore.

PR #2035, opened 2022-09-16
Fix incorrect argc value in run_vy_app.cc

siyuan0322:


analytical_engine/test/run_vy_app.cc
```diff
@@ -491,7 +491,7 @@ void Run(vineyard::Client& client, const grape::CommSpec& comm_spec,
 }
 
 int main(int argc, char** argv) {
-  if (argc < 9) {
+  if (argc < 8) {
```
sighingnow, line 494: 8 is incorrect as well, image imaging we have 1 edge, 1 vertex, and no app name as well as no path pattern is given.

siyuan0322: In this case it should report error?
Now the minimum parameters are `./run_vy_app <ipc_socket> <e_label_num> <efiles...> "
        "<v_label_num> <vfiles...> <run_projected>`

siyuan0322, line 494:
```suggestion
  if (argc < 6) {
```

PR #2038, opened 2022-09-16
try to fix grpc protobuf version

zhanglei1949:
## Related issue number

Ensuring compatibility between python package.`grpcio` and `protobuf`.

Not sure about whether this PR will FIX...

### DIFF
--- a/coordinator/setup.py
+++ b/coordinator/setup.py
@@ -374,8 +374,9 @@ setup(
     },
     setup_requires=[
         "setuptools_scm>=5.0.0",
-        "grpcio<=1.43.0,>=1.40.0",
-        "grpcio-tools<=1.43.0,>=1.40.0",
+        "grpcio",
+        "grpcio-tools",
+        "protobuf>=3.20.0,<4",
     ],
     package_dir=parsed_package_dir(),
     packages=parsed_packages(),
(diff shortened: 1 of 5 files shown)


coordinator/setup.py
```diff
@@ -374,8 +374,9 @@ def parse_version(root, **kwargs):
     },
     setup_requires=[
         "setuptools_scm>=5.0.0",
-        "grpcio<=1.43.0,>=1.40.0",
-        "grpcio-tools<=1.43.0,>=1.40.0",
+        "grpcio",
+        "grpcio-tools",
```
sighingnow, line 378: If there's no version control for grpcio and grpcio-tools, the build process may failure in user's development environemnt as well, (imaging there's are some staled version installed)

python/setup.py
```diff
@@ -289,8 +289,9 @@ def parse_version(root, **kwargs):
     },
     setup_requires=[
         "setuptools_scm>=5.0.0",
-        "grpcio<=1.43.0,>=1.40.0",
-        "grpcio-tools<=1.43.0,>=1.40.0",
+        "grpcio",
+        "grpcio-tools",
+        "protobuf>=3.20.0,<4",
```
sighingnow, line 294: Ditto.

zhanglei1949:
Closed since as there is no need to update these two package to a new version.

PR #2057, opened 2022-09-21
[Bug Fix] fix bug in label encoding of `exp_store`

BingqingLyu:
## What do these changes do?

As titled.

## Related issue number

#2013


BingqingLyu:
This bug would cause unexpected results for some bi queries. Here, we fix label encoding in `GraphProxy`. In the future, it would be fixed in `ExpStore`.

longbinlai (approved):
Okay for now. But ultimately we should modify the implementation of exp_store.

PR #2058, opened 2022-09-21
[GIE-IR] fix expand fusion strategy

shirly121:
## What do these changes do?
fuse bothE() with otherV instead of bothV()

## Related issue number

Fixes


longbinlai (approved):
LGTM.

PR #2076, opened 2022-09-26
[GIE-IR] support sending back a page of results if batch size met

shirly121:
## What do these changes do?
To support result paging in GraphScope Studio, compiler will send back results in a batch way instead of sending all by once

## Related issue number

Fixes


siyuan0322 (approved):
👍

PR #2088, opened 2022-09-28
Closing the session if hearbeat fails for many times (when engine crashes)

sighingnow:
## What do these changes do?

It is quite strange to keep printing messages / exceptions like heartbeat failures when engine crashes. In this pull request, the session will be automatically closed after `self._heartbeat_maximum_failures` retires. Note that `self._heartbeat_maximum_failures` is customizable so it should fit advanced users well.

This pull reqests also rename the dummy package name "foo" to "graphscope_runtime".

## Related issue number

N/A


coordinator/setup.py
```diff
@@ -288,7 +288,7 @@ def parsed_packages():
     name = os.environ.get("package_name", "gs-coordinator")
     if name == "gs-coordinator":
         return find_packages(".")
-    return ["foo"]
+    return ["graphscope_runtime"]
```
lidongze0629, line 291: Also change the  module name 'foo' in coordinator directory

https://github.com/alibaba/GraphScope/tree/main/coordinator/foo

siyuan0322 (approved):

coordinator/setup.py
sighingnow: Already renamed.

PR #2102, opened 2022-09-29
[GIE/IR] Fix the bug when fuse filter in `Auxilia`

### LINKED ISSUE #2021: [BUG] Fuse `filter` into `Auxilia` in `GIE-IR` occurs bugs
**Describe the bug**

In `GIE-IR`, it would fuse `filter` into `Auxilia` to enable optimization of pushing filter down into Store (or StoreProxy). However, some filters cannot be push down, since it may involve computed data.
An example is shown as follows (the query is meaningless):
`g.V().hasLabel("PERSON").as("a").limit(10).count().as("c").select("a").out("KNOWS").where(expr("@c<@.creationDate"))`,
which outputs empty results.

BingqingLyu:
## What do these changes do?

As titled. Currently, in `Auxilia`, we will try to evaluate the filter expression before pushing down to the storage, in case that some computed values of variables are missing. Specially, if all the values of variables can be pre-fetched, we will evaluate directly without pushdown the filter.

## Related issue number

Fixes #2021


longbinlai:
Should be further designed

PR #2112, opened 2022-10-09
[Pegasus] Avoid get iter.next() when it is None already in sending batches

### LINKED ISSUE #2114: [BUG] unexpected behavior found in EdgeExpand in queries
**Describe the bug**

e.g., in query `g.V().hasLabel('person').out('created')`, we have results of:

v4->created->v3,
v4->created->v5,
v6->created->v3,
...

The expected behavior is that we iterate over outgoing neighbors of v4, getting results of {v3, v5} and `None`(indicating no more neighbors), and then continue to find neighbors of other vertices (e.g., v6).

However, when iterating over neighbors of v4, we actually get `None` **TWICE** before processing v6, which is not as expected.

BingqingLyu:
## What do these changes do?

As titled. 

## Related issue number

Fixes #2114

### DIFF
--- a/interactive_engine/executor/engine/pegasus/pegasus/src/communication/output/output.rs
+++ b/interactive_engine/executor/engine/pegasus/pegasus/src/communication/output/output.rs
@@ -399,7 +399,7 @@ impl<D: Data> OutputHandle<D> {
                 }
                 Ok(None) => {
                     // all data in iter should be send;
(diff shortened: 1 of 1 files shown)


longbinlai (commented):
Look good

longbinlai (approved):
Look good

PR #2123, opened 2022-10-10
[GAE-Java] GraphX-on-GraphScope : remove `GraphXFragment` and related components

zhanglei1949:
### Background

The original GraphX-on-GraphScope implementation support both `GraphXFragment` And `ArrowProjectedFragment`. Relying on two storage structure introducing unnecessary  complexity and GraphXFragment doesn't differ much from `ArrowProjectedFragment` from a design point of view . So in this PR, we remove GraphXFragment and related components, and refactor Graphx-on-GraphScope to adapt this change.

### Changes.
- Remove all GraphXFragment related stuff in c++ and java


analytical_engine/java/grape-runtime/src/main/java/com/alibaba/graphscope/annotation/AnnotationInvoker.java
```diff
@@ -1341,7 +1341,7 @@
                                                 CPP_ARROW_PROJECTED_FRAGMENT
                                                         + "<int64_t,uint64_t,double,int64_t>",
                                                 "double",
-                                                "double"
+                                                "whatever"
```
sighingnow, line 1344: `any` might be better than `whatever` for data types.

luoxiaojian (approved):

PR #2130, opened 2022-10-12
App now can carry cmake options, and enabled wcc on string oid.

siyuan0322:
And fixed errors in precompile step.
https://github.com/alibaba/GraphScope/actions/runs/3221735048/jobs/5270025612

WCC on string will use global ID as component ID.


sighingnow (approved):
LGTM.

Seems there are still some temporary comment blocks.

k8s/precompile.py
```diff
@@ -380,5 +398,5 @@ def compile_cpp_pie_app():
 
 if __name__ == "__main__":
     os.makedirs(WORKSPACE, exist_ok=True)
-    compile_graph()
+    # compile_graph()
```
lidongze0629, line 401: Maybe you forget to comment out this line

PR #2139, opened 2022-10-17
Update cluster.py

changyi1223:
fix the bug that etcd selector could not work

## What do these changes do?

we have test the k8s_label_selector ，the labels on coordinate and engine work well ,but etcd doesn't. We followed the code and finally located here, then fixed it to make it work.

## Related issue number

Fixes

### DIFF
--- a/coordinator/gscoordinator/cluster.py
+++ b/coordinator/gscoordinator/cluster.py
@@ -749,7 +749,6 @@ class KubernetesClusterLauncher(Launcher):
         )
         if self._etcd_pod_node_selector:
             etcd_builder.add_etcd_pod_node_selector(self._etcd_pod_node_selector)
-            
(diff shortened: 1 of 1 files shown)


sighingnow (approved):
Nice catch!

Thanks!

coordinator/gscoordinator/cluster.py
```diff
@@ -747,7 +747,9 @@ def _create_etcd(self):
             listen_peer_service_port=self._etcd_listening_peer_port,
             listen_client_service_port=self._etcd_listening_client_port,
         )
-
+        if self._etcd_pod_node_selector:
+            etcd_builder.add_etcd_pod_node_selector(self._etcd_pod_node_selector)
+
```
sighingnow, line 752:
```suggestion

```

PR #2152, opened 2022-10-20
Reorganized coordinator and redesigned the logic to create engines.

### LINKED ISSUE #2019: Refactor coordinator and split it to modules
The coordinator has multiple non-related functionalities, and we should break it up into small modules and make it more clear.

### LINKED ISSUE #2020: Layering of docker images.
- [x] GAE
- [x] GIE
- [x] GLE
- [x] Coordinator
- [x] Groot
- [x] Runtime

### LINKED ISSUE #2155: Confoming to common practice of make and make install in the top level Makefile
- [x] Currently, the make graphscope will also install artifacts to the directory, specified by `INSTALL_PREFIX`, which it shouldn't.
- [x] The gae, gie and gle is a PHONY target, which means it will be rebuilded from scratch every time when hitting `make`

### LINKED ISSUE #2156: Make groot not depend on vineyard
groot has nothing to do with vineyard, but it will complain when it cannot find vineyard.

The reason is the rust package depend on v6d, and which is unnecessary and should be removed.

siyuan0322:
Coordinator:
1. Split coordinator to launcher and op_executor
2. Revised the logic for connecting session
3. Create and close engines now use their own RPC call instead of Operation in DAG
4. Simplify and refine error handling
5. Organized local launcher and kubernetes launcher
6. Move the create gl handle parts to client, in order to keep coordinator simple

Interactive engine:
1. Separate groot and v6d rust package compilation as they are non-related
2. Thanks to the separation of compilation and the dependency reduction, the image size shrinks from 7GB to 0.9GB
7. Clean more target in `make clean`

Client:
1. Refactored session and RPC client according to changes in coordinator
2. Removed some unused parameters in functions
3. Creating GIE and GLE now are not parts of DAG, hence their DAGNode are removed
4. Remove some unused parameters

Dev infra:
1. Use NUMPROC to control compilation parallelism instead of hard coded value
2. Eliminate usage of sudo inside Makefile
3. Make now only do make in local, and make install will copy it to destination directory
4. Make GAE, GIE, and GLE now will use cached artifact instead of build from beginning
5. Add a bunch of dockerfiles for single modules (GAE, GIE, GLE, Coordinator).

Miscellaneous:
1. Fix several typos

Fixes #2020 
Fixes #2019 
Fixes #2155
Fixes #2156


sighingnow (approved):

Makefile
```diff
@@ -1,178 +1,179 @@
...
+
+graphscope-store-image:
+	$(MAKE) -C $(K8S_DIR) graphscope-store-image VERSION=$(VERSION)
+
+push:
+	$(MAKE) -C $(K8S_DIR) push
+
+
+## Tests
+.PHONY: test, unittest, minitest, k8stest
```
sighingnow, line 163: No comma between `.PHONY` dependent targets in Makefile.

PR #2205, opened 2022-11-04
[Bug Fix] Add latest tag to all image in graphscope's helm chart

dashanji:
While installing the graphscope helm chart, as the tag is nil, the image is rendered as `registry.cn-hongkong.aliyuncs.com/graphscope/graphscope:` so we can't install it as expected.

### DIFF
--- a/charts/graphscope-store/templates/_helpers.tpl
+++ b/charts/graphscope-store/templates/_helpers.tpl
@@ -70,12 +70,12 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 Return the proper graphscope-store image name
 */}}
 {{- define "graphscope-store.image" -}}
-{{ include "graphscope-store.images.image" (dict "imageRoot" .Values.image "global" .Values.global) }}
+{{ include "graphscope-store.images.image" (dict "imageRoot" .Values.image "global" .Values.global "DefaultTag" .Chart.AppVersion ) }}
 {{- end -}}
 
 {{/*
 Return the proper image name
-{{ include "graphscope-store.images.image" ( dict "imageRoot" .Values.path.to.the.image "global" $) }}
+{{ include "graphscope-store.images.image" ( dict "imageRoot" .Values.path.to.the.image "global" $ "DefaultTag" .Chart.AppVersion ) }}
 */}}
 {{- define "graphscope-store.images.image" -}}
 {{- $registryName := .imageRoot.registry -}}
@@ -85,6 +85,13 @@ Return the proper image name
     {{- if .global.imageRegistry }}
      {{- $registryName = .global.imageRegistry -}}
     {{- end -}}
+{{- end -}}
+{{- if not $tag }}
+{{- if .imageRoot.DefaultTag }}
+{{- $tag = .imageRoot.DefaultTag -}}
+{{- else -}}
+{{- $tag = "latest" -}}
+{{- end -}}
 {{- end -}}
 {{- if $registryName }}
 {{- printf "%s/%s:%s" $registryName $repositoryName $tag -}}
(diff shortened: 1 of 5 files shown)


sighingnow:
The image tag was expected to be `Chart.AppVersion`: https://github.com/alibaba/GraphScope/blob/main/charts/graphscope/templates/coordinator.yaml#L22

Don't use latest as tag please as it hard to ensure compatibility with the latest image version and the old chart versions.

sighingnow:
If `AppVersion` is not rendered correctly, I suggest you figure out why it is not rendered, rather than changing the default value to "latest".

dashanji:
Make sense, the latest commit is to add `default tag(Chart.AppVersion)` to graphscope-store's helm chart. 

It fix the following error:
```shell
$ helm install graphscope-store "./graphscope-store"
Error: INSTALLATION FAILED: YAML parse error on graphscope-store/templates/coordinator/statefulset.yaml: error converting YAML to JSON: yaml: line 39: mapping values are not allowed in this context
```

sighingnow (approved):

charts/graphscope-store/templates/_helpers.tpl
```diff
@@ -86,6 +86,13 @@ Return the proper image name
      {{- $registryName = .global.imageRegistry -}}
     {{- end -}}
 {{- end -}}
+{{- if not $tag }}
+{{- if .imageRoot.DefaultTag }}
```
siyuan0322, line 90: It seems the new kv you added isn't used ("DefaultTag" .Chart.AppVersion)?
The `.imageRoot.DefaultTag` points to the `image: DefaultTag` in values.yaml?
So I guess if you install the charts without tag specified, you will have `latest` tag instead of `0.17.0` or `0.18.0` (the AppVersion)

siyuan0322: Maybe should change the `.imageRoot.DefaultTag` to `.DefaultTag`?

charts/graphscope-store/templates/_helpers.tpl
```diff
@@ -70,12 +70,12 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 Return the proper graphscope-store image name
 */}}
 {{- define "graphscope-store.image" -}}
-{{ include "graphscope-store.images.image" (dict "imageRoot" .Values.image "global" .Values.global) }}
+{{ include "graphscope-store.images.image" (dict "imageRoot" .Values.image "global" .Values.global "DefaultTag" .Chart.AppVersion ) }}
```
dashanji, line 73: @siyuan0322 The new kv is contained in the dict, and I think the `imageRoot.DefaultTag` points to the `.Chart.AppVersion` of the `Chart.yaml`. Is there something I missed?

In fact, we can get the following result.
```shell
$ helm install graphscope-store "./graphscope-store"
$ kubectl get po coordinator-graphscope-6d8fccbb9c-cl6vm -oyaml | grep image
    - --k8s_gs_image
    - --k8s_etcd_image
    - --k8s_image_pull_policy
    - --k8s_image_pull_secrets
    image: registry.cn-hongkong.aliyuncs.com/graphscope/graphscope:0.18.0
    imagePullPolicy: Always
  - image: registry.cn-hongkong.aliyuncs.com/graphscope/jupyter:0.18.0
    imagePullPolicy: Always
```

siyuan0322 (approved):

charts/graphscope-store/templates/_helpers.tpl
dashanji: Thanks for the catching, I have tried again and find the tag is exactly `latest`.

```shell
$ kubectl get po graphscope-store-coordinator-0 -n gh-store -oyaml | grep image
            f:image: {}
            f:imagePullPolicy: {}
    image: registry.cn-hongkong.aliyuncs.com/graphscope/graphscope-store:latest
    imagePullPolicy: IfNotPresent
    image: registry.cn-hongkong.aliyuncs.com/graphscope/graphscope-store:latest
    imageID: registry.cn-hongkong.aliyuncs.com/graphscope/graphscope-store@sha256:b00a5a15a48a5bca86d02f7c16a07335377bc3e617b3d67c22331b5a5ba9008a
```

I'll fix it ASAP.

PR #2246, opened 2022-11-25
Reorganize for Makefiles and dockerfiles

siyuan0322:
## What do these changes do?

This PR reorganized Makefiles and dockerfiles, targets for the consistency of image and non-image version and ease of management.

Some major changes includes:

- Redesigned the directory strucure under `k8s`, make similar targets appears in the same directory. 
    - Some internal stuffs, which very unlikely would be used and be cared about  by users, such as building wheels, building dataset image go to the `internal` directory
    - All dockerfiles go to the `dockerfiles` directory
    - Building scripts go to the `build_scripts` directory
    - All other scripts go to the `utils` directory
    - Add documentation.
- Unified the make targets in root directory and k8s directory. Ideally, the only difference should be that one would produce local artifacts and the other would produce the image.
- Drop unnecessary make targets - push, clean
- Drop registry fields in developing stage for better readability
- Disable the release of nightly jupyter, dataset and graphscope-store image, it's not yet necessary.

## Related issue number

Fixes N/A


sighingnow (approved):
Looks like there still are many temporay comments, you may need to do a cleanup before merging.

sighingnow:
> > Looks like there still are many temporary comments, you may need to do a cleanup before merging
> 
> Do you mean those in CI yamls?

Yes.

PR #2291, opened 2022-12-05
Move the theme to furo, and add some markdown examples.

sighingnow:
## What do these changes do?

as titled.

## Related issue number

Fixes N/A


sighingnow:
It shouldn't affect current "stable" documentation (a.k.a. https://graphscope.io/docs) and should be safe to merge.

PR #2313, opened 2022-12-06
Vineyard use local meta instead of etcd in hosts mode

siyuan0322:
In hosts (local) mode, etcd is not a strong dependency for vineyard, and we could leverage vineyard's local meta to simplify the starting process.

The etcd stuffs are not dropped, as it's still used in Kubernetes.
I'll consider droping the unused miscellaneous etcd codes in another deployment PR.

### DIFF
--- a/coordinator/gscoordinator/local_launcher.py
+++ b/coordinator/gscoordinator/local_launcher.py
@@ -92,7 +92,7 @@ class LocalLauncher(AbstractLauncher):
     def stop(self, is_dangling=False):
         self.close_analytical_instance()
         self.close_vineyard()
-        self.close_etcd()
+        # self.close_etcd()
 
     def set_session_workspace(self, session_id):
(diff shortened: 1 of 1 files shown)


siyuan0322:
This is a proposal, since it's very lightweight so I just PR it.
Feel free to reject this if not appropriate.

sighingnow:
As far as I know, when `cluster_type = "hosts"`, the `hosts` argument can be a list of nodes (to launch MPI processes on multiple workers) and when such settings  is applied, etcd is required.

AFAIK, such a feature is indeed used somewhere.

siyuan0322:
Okay~

sighingnow:
Yes, but may be every helpful.

Later GART will introduce hard dependency for etcd for metadata synchronization, even on a single node.

siyuan0322:
> Later GART will introduce hard dependency for etcd for metadata synchronization, even on a single node.

It's okay, we could enable etcd when carried with gart only~

PR #2324, opened 2022-12-09
Fix helm could not delete role and rolebindings when uninstalling GS

siyuan0322:
Helm uninstall would let role and rolebinding of GraphScope remains.
- In #2208 we use `patch` to delete `role` and `rolebindind`s, but `patch` seems would depends `get`
- Previous `role_and_rolebinding.yml` is no longer used.

After `get` is added, all resources could be deleted.


sighingnow:
Why mixing totally unrelated changes to this pr ...

PR #2330, opened 2022-12-14
[GAE-Java] Pack JNI lib `libgrape-jni.so` into `grape-runtime.jar`

zhanglei1949:
- Pack `libgrape-jni.so` into `grape-runtime-shaded.jar`.
- Remove target library name in `@FFIGen("grape-jni")`, to avoid FastFFI generating `System.loadLibrary()` code.
- Strip jni library after building, shrinking size from 500MB to <10MB
- Fix symbol look up issue in codegen&built app as jni lib is striped.


sighingnow (approved):

analytical_engine/java/grape-runtime/src/main/java/com/alibaba/graphscope/runtime/LoadLibrary.java
```diff
@@ -30,7 +32,11 @@ public class LoadLibrary {
     private static Logger logger = LoggerFactory.getLogger(LoadLibrary.class);
 
     static {
-        System.loadLibrary("grape-jni");
+        try {
+            NativeLoader.loadLibrary("grape-jni");
+        } catch (IOException e) {
+            e.printStackTrace();
```
siyuan0322, line 38: I wonder if the exception got swallowed is by purpose?

siyuan0322 (approved):

analytical_engine/java/grape-runtime/src/main/java/com/alibaba/graphscope/runtime/LoadLibrary.java
zhanglei1949: Yes, we try to load jni library, but it is ok if failed.

PR #2344, opened 2022-12-23
[Bug fix] fix delete fragment failed when unloading graph

songqing:
## What do these changes do?

When unloading graph, the current logic is deleting fragment group first and then deleting the corresponding fragments.
However, deleting fragment group will also deleting the corresponding fragments automatically as we use the param "deep", then when we delete the fragments of the group, they don't exist and that will cause core dump when compiling graph_engine with "Debug" option.
g2

So, remove the logic of deleting fragments to solve the problem.

## Related issue number

This bug can be recurrent when running the python unittest  "project_to_simple_with_name" in "tests/unittest/test_graph.py"

### DIFF
--- a/analytical_engine/core/grape_instance.cc
+++ b/analytical_engine/core/grape_instance.cc
@@ -176,6 +176,7 @@ bl::result<rpc::graph::GraphDefPb> GrapeInstance::loadGraph(
 
 bl::result<void> GrapeInstance::unloadGraph(const rpc::GSParams& params) {
   BOOST_LEAF_AUTO(graph_name, params.Get<std::string>(rpc::GRAPH_NAME));
+  VLOG(1) << "Unloading Graph " << graph_name;
   if (params.HasKey(rpc::VINEYARD_ID)) {
     BOOST_LEAF_AUTO(frag_group_id, params.Get<int64_t>(rpc::VINEYARD_ID));
     bool exists = false;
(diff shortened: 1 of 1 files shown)


sighingnow:
Hi @songqing,

Thanks for raising this issue up!

Which version of graphscope you are working with?

sighingnow:
I believe the issue has already been fixed in vineyard in https://github.com/v6d-io/v6d/pull/1067, which, has already been included in vineyard>=v0.11.0.

We encourage to upgrade vineyard to either v0.11.0 or v0.11.1 for the fixes. If you are working with "main" branch of graphscope, just install our latest vineyard release v0.11.3.

Thanks!

songqing:
> 

The GraphScope version I used is 0.17.0.

songqing:
OK, the fix in v6d may not cause the graph_engine process core dump again when unloading graph, but the problem I mentioned still exists.

sighingnow:
The duplicated call for "DelData" is to guarantee the graph been dropped when unloading graph. We did have encountered some corner cases where the graph is failed to remove from memory if we only delete "frag_group_id".

Ideally `DelData()` on an non-exist object should casuing nothing.

songqing:
ok, I see, how about adding a check if the fragment exist, then delete it?

sighingnow:
> ok, I see, how about adding a check if the fragment exist, then delete it?

DelData = if exists then delete else noop

songqing:
> > ok, I see, how about adding a check if the fragment exist, then delete it?
> 
> DelData = if exists then delete else noop

OK, that's good, I'll close this PR, thank you.

sighingnow:
> OK, that's good, I'll close this PR, thank you.

Thanks for being interested in GraphScope! Feel free to open issues or pull requests if you have any further question!

sighingnow:
Closing, as discussed.

PR #2345, opened 2022-12-25
Fix GAE v6d test with arrow projected fragment

songqing:
## What do these changes do?

The run_vy_app.cc contains multiple algorithm tests, and some of them are about ArrowProjectedFragment.
However, now all these algorithms use the same params with the Project(), that is "ProjectedFragmentType::Project(fragment, 0, 0, 0, 0)", the last param of this function is "e_prop_id".
For SSSP, we need edge data, and the e_prop_id is not 0(0 is the src id column), but 2, so, the result of SSSP is wrong, see the following pic.
g3

## Related issue number

Fixes
So, we need multiple ProjectedFragmentTypes with different node end edge types, and that's the PR does.


sighingnow (approved):

sighingnow:
LGTM.

songqing:
@sighingnow help to approval the workflows, thanks

PR #2356, opened 2022-12-27
Add ownerreferences to all subresources created by coordinator

dashanji:
## What do these changes do?

The previous method was too tricky via https://github.com/alibaba/GraphScope/pull/2350, and I think the recommended way is adding ownerreferences to add subresources created by the coordinator.


sighingnow (approved):

siyuan0322 (approved):

sighingnow:
It should be ok to merge and the CI failure would been handled in #2354.

PR #2360, opened 2022-12-28
Fix missing header files

songqing:
## What do these changes do?

When building my test, I encounter some compile errors as some header files are missed, besides, namespace is not assigned in some macro definitions.

## Related issue number

Fixes


sighingnow (approved):
LGTM. Thanks!

PR #2402, opened 2023-02-01
[BUG Fix] fix bug when fetch properties of tags generated by a `Project`

### LINKED ISSUE #2401: [BUG] Cannot output expected results in some cases of `Project`
**Describe the bug**

Some properties projection queries outputs unexpected results.

e.g., run the following query on modern graph:
`g.V().out().as("a").in().select("a").as("b").select("b").by("name")`;

It outputs empty results, while the expected results is:

```
gremlin>  g.V().out().as("a").in().select("a").as("b").select("b").by("name")
==>lop
==>lop
==>lop
==>vadas
==>josh
==>lop
==>lop
==>lop
==>ripple
==>lop
==>lop
==>lop
```

In this case, when project tag `a` as `b` (like a tag renaming), we cannot  further get properties of `b`, which is not as expected.

BingqingLyu:
## What do these changes do?

1. Fix the bug when we want to fetch properties of tags that generated by `Project` opr,
e.g., `g.V().out().as("a").in().select("a").as("b").select("b").by("name")`
2. Further replace `Auxilia` (to remove unnecessary system-given tags) by `Project` (to project necessary user-given tags) in `Pattern`

## Related issue number

Fixes #2401

### DIFF
--- a/interactive_engine/compiler/src/test/java/com/alibaba/graphscope/integration/IrGremlinQueryTest.java
+++ b/interactive_engine/compiler/src/test/java/com/alibaba/graphscope/integration/IrGremlinQueryTest.java
@@ -42,6 +42,8 @@ public abstract class IrGremlinQueryTest extends AbstractGremlinProcessTest {
 
     public abstract Traversal<Vertex, Object> get_g_V_hasLabel_hasId_values();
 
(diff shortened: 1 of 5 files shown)


longbinlai (approved):
LGTM

PR #2455, opened 2023-02-23
Fixes building wheel process on local CI.

sighingnow:
## What do these changes do?

It have been failed for quite a while... Removing unsupported Python 3.7 from CI (and install_deps.sh) as well.


sighingnow:
See  also https://github.com/alibaba/GraphScope/actions/runs/4249240830/jobs/7389565466, the local ci has been install incorrect artifacts since many revisions ago ...

PR #2459, opened 2023-02-24
[doc] Enable colon-fence extension by default and use better image syntax.

sighingnow:


sighingnow:
Merging as works as expected.

PR #2475, opened 2023-03-01
Fix port of learning container is not exposed error

siyuan0322:


LiSu (approved):

coordinator/gscoordinator/cluster_builder.py
```diff
@@ -315,6 +315,9 @@ def get_learning_container(self, volume_mounts):
             self._learning_requests,
             self._learning_requests,
         )
+        container.ports = [
+            kube_client.V1ContainerPort(container_port=p) for p in range(8000, 9000)
+        ]
```
sighingnow, line 320: Respect `_learning_start_port` please, don't hard code the port range.

Issue #2524, opened 2023-03-15
[BUG] Gremlin startup error when k8s release name contains' - '.


YANGBoSunning:
**Describe the bug**
Gremlin startup error when k8s release name contains' - '. 

**To Reproduce**
1. Pass the name of release like 'a-b' in the helm install command
2. Create a session and start the engine
3. Execute code 'sess.gremlin(g)'
4. Error occurred

**Expected behavior**

**Screenshots**

**Environment (please complete the following information):**
 - GraphScope version: v0.19.0
 - OS: Linux
 - Kubernetes Version: 1.22.10

PR #2526, opened 2023-03-15
Fix: Gremlin startup error when k8s release name contains' - '. #2524

YANGBoSunning:
## What do these changes do?

When starting the gremlin instance, you need to get the instance_id. If the release name of the current application contains' - ', for example, 'b-c', the coordinator_name is' coordinator-b-c ', and every char after the last '-' is captured in the case of existing code, and the instance_id is' c '. In fact, every char after the first '-' should be captured. instance_id is' b-c '.

## Related issue number

#2524 

Fixes


siyuan0322 (approved):
Nice catch! Indeed it is.
Thanks.

Issue #2528, opened 2023-03-16
Add a ci step to avoid tmate operations on self-hosted machines


dashanji:
The feature is to avoid data leaks on self-hosted machines.

PR #2529, opened 2023-03-16
Add a CI step to avoid running the tmate session on the self-hosted machine

dashanji:
## What do these changes do?

As titled.

## Related issue number

Fix https://github.com/alibaba/GraphScope/issues/2528.

### DIFF
--- a/.github/workflows/gaia.yml
+++ b/.github/workflows/gaia.yml
@@ -27,6 +27,13 @@ jobs:
       with:
         submodules: true
 
+    - name: Detect the tmate session
+      run: |
+        if grep -v "grep" .github/workflows/gaia.yml | grep "action-tmate"; then
+          echo 'WARNING!!!the self-hosted machine can not run tmate session, please debug it manually'
+          exit 1
+        fi
+
     - uses: actions/cache@v3
       with:
         path: ~/.m2/repository
--- a/.github/workflows/pegasus.yml
+++ b/.github/workflows/pegasus.yml
@@ -52,6 +52,13 @@ jobs:
     steps:
     - uses: actions/checkout@v3
 
+    - name: Detect the tmate session
+      run: |
+        if grep -v "grep" .github/workflows/pegasus.yml | grep "action-tmate"; then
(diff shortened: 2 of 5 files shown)


sighingnow (approved):
LGTM.

dashanji:
Wait for testing here.

.github/workflows/gaia.yml
```diff
@@ -23,6 +23,12 @@ jobs:
   gaia-test:
     runs-on: [self-hosted, manylinux2014]
     steps:
+    - name: Detect the tmate session
+      if: "jobs.gaia-test.steps[*].uses == 'mxschmitt/action-tmate@*'"
```
siyuan0322, line 27: could we use '*/action-tmate'?
Since mxschmitt is not the only one that provides action-tmate.

siyuan0322 (approved):

.github/workflows/gaia.yml
dashanji: Make sense to me.

dashanji:
From the [github steps context](https://docs.github.com/en/actions/learn-github-actions/contexts#steps-context), we can't get the info of steps that wait for running. So use a static code check here.

.github/workflows/pegasus.yml
```diff
@@ -50,6 +50,12 @@ jobs:
   k8s-test:
     runs-on: [self-hosted, ubuntu2004]
     steps:
+    - name: Detect the tmate session
+      if: "jobs.k8s-test.steps[*].uses == 'mxschmitt/action-tmate@*'"
```
siyuan0322, line 54: why this is still use the context while others use `grep`

sighingnow (commented):
`k8s-ci.yaml` and `local-ci.yaml`: put the check to `changes` job as others will depends on that to avoid duplication so many times..

.github/workflows/pegasus.yml
dashanji: My bad. Thanks for catching this.

dashanji:
> `k8s-ci.yaml` and `local-ci.yaml`: put the check to `changes` job as others will depends on that to avoid duplication so many times..

Addressed.

PR #2534, opened 2023-03-20
[Bug Fix] Fix bug in getting properties with `LateProject` strategy 

### LINKED ISSUE #2535: [BUG] get properties failed in GIE with `LateProject` strategy
**Describe the bug**
With `LateProject` strategy, the following queries output empty results:
1. `g.V().out().values("age")` with worker_num=1, and
2. `g.V().out().valueMap()`

BingqingLyu:
## What do these changes do?

1. Fix bug in getting properties with LateProject strategy by supporting get_property  (e.g., when for `IdOnlyVertex`), and get_all_property from storage directly when necessary.
2. Modify the logic in `LazyDetails`, that when get_all_property(), we do get all properties from the store vertex, instead of getting all `pre-cached` properties from the runtime vertex. However, when encoding the `LazyDetails`, it will carry the `pre-cached` properties, since we want to visit these properties on remote partitions.
3. Remove hybrid_details when GIE on Groot, as it is not necessary in `LateProject` strategy.

## Related issue number

Fixes #2535


longbinlai (approved):
LGTM

PR #2541, opened 2023-03-20
Add document tutorial_dev_algo_cpp_pie.md

lidongze0629:
## What do these changes do?

- [x] tutorial_dev_algo_cpp_pie.md
- [ ] how_to_find_logs.md

## Related issue number

Fixes

### DIFF
--- a/docs/analytical_engine/tutorial_dev_algo_cpp_pie.md
+++ b/docs/analytical_engine/tutorial_dev_algo_cpp_pie.md
@@ -1,2 +1,148 @@
 # Tutorial: Develop your Algorithm in C++ with PIE Model
 
+In this tutorial, we will demonstrate how to develop an algorithm using the PIE model in C++ programming language.
+
+## Prerequisites:
+- Familiarity with the [PIE programming model](https://graphscope.io/docs/latest/analytical_engine/programming_model_pie.html) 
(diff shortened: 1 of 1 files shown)


docs/analytical_engine/tutorial_dev_algo_cpp_pie.md
```diff
@@ -1,2 +1,148 @@
...
+    result.SetValue(0);
+  }
+
+  int step = 0;
+  int param1 = 0;
+  typename FRAG_T::template vertex_array_t<uint64_t>& result;
+};
+```
+
+As shown in the code. Respectively, The `MyAppContext` class defines two member variables called `step` and `param1` to store the current superstep and algorithm-specific parameter. And also defines a member variable named `result` with `uint64_t` type to store the the degree for each vertex in the fragment. The `Init` method is used to initialize the context of the computation. In current example, we initialize the `step` and `param1` variables to zero and the algorithm-specific parameter. We also set the result to zero for each vertex.
```
doudoubobo, line 39: As shown in the code, the `MyAppContext` class defines two member variables called `step` and `param1` to store the current superstep and algorithm-specific parameter, respectively. And we also define a member variable named `result` with `uint64_t` type to store the the degree for each vertex in the fragment.

doudoubobo (approved):

PR #2681, opened 2023-05-09
[GIE/Runtime] Remove the dependency of GIE Runtime on GlobalQueryStore to accelerate the compilation

### LINKED ISSUE #2679: Remove the dependency of GIE Runtime on GlobalQueryStore to accelerate the compilation

Remove the dependency of GIE Runtime on GlobalQueryStore. Then GIE won't depend on RocksDB (introduced by GlobalQueryStore), which will accelerate GIE compilation.

BingqingLyu:
## What do these changes do?

 As titled.

Notice that by default compiling (`cargo build` in ir-core), it won't depend on `GlobalQuery` anymore. If you want to compile GIE based on `GlobalQuery` (including Groot and V6d), do `cargo build --features with_global_query`.

## Related issue number

Fixes #2679

### DIFF
--- a/interactive_engine/compiler/Makefile
+++ b/interactive_engine/compiler/Makefile
@@ -31,6 +31,7 @@ clean:
 test:
 	cd $(CUR_DIR)/../executor/ir && cargo test && \
 	cd $(CUR_DIR)/../executor/ir && cargo test --features with_v6d && \
+	cd $(CUR_DIR)/../executor/ir && cargo test --features with_global_query && \
 	cd $(CUR_DIR) && mvn test
 
 # start rpc server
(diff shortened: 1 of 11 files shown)


siyuan0322 (approved):
Seems this PR only takes effect when compile ir on experimental store.

So if we do `make interactive` it still need to compile rocksdb twice..

Now I understand it's because the global_query is a interface like crate, which depends on rocksdb, which is not a decent design. Someday we may want to drop this dependency.

BingqingLyu:
> 

Agreed. I think this dependency (GlobalQuery based on RocksDb) can be removed once we introduced GRIN.

BingqingLyu:
> Seems this PR only takes effect when compile ir on experimental store.
> 
> So if we do `make interactive` it still need to compile rocksdb twice..
> 
> Now I understand it's because the global_query is a interface like crate, which depends on rocksdb, which is not a decent design. Someday we may want to drop this dependency.

Agreed. This is expected to be resolved after we introduced GRIN.

PR #2683, opened 2023-05-09
Export the k8s logs of graphscope helm test and upload to artifact when the helm test failed

dashanji:
## What do these changes do?

As titled.


dashanji:
I'm sorry we can't dump the logs of the k8s-ci test at present, and current `kubernetes-export-logs` can only support dumping all logs at a time rather than a period of time. As all k8s resources are managed by the coordinator, and whether the pytest succeeded or failed, the k8s resources will be deleted after that.

siyuan0322:
gss.yaml also has a helm test.

dashanji:
Added. @siyuan0322, please take a look at this.

PR #2744, opened 2023-05-25
[GIE/Runtime] Redesign `PartitionerInfo`, `ClusterInfo`, and `Router` trait to better support parallel processing in Runtime

### LINKED ISSUE #2753: [GIE] Refine `Partitioner` trait to better support parallel scan

Refine `Partitioner` trait to better support parallel scan. 

Currently, `Partitioner` define the `get_worker_partitions` function to specify the partition list that the worker (i.e., threads) is going to process:
```
 fn get_worker_partitions(&self, job_workers: usize, worker_id: u32) -> GraphProxyResult<Option<Vec<u64>>>;
```
However, we may have `m` workers and `n` partitions, where `m>n`. Thus, each worker need to scan only part of the partition. Take this case into consideration and refine the trait.

BingqingLyu:
## What do these changes do?

Redesign `PartitionInfo`, `ClusterInfo`, and `Router` trait to better support parallel processing in Runtime, where:
* `PartitionInfo` is used to query the partition information when the data has been partitioned.
* `ClusterInfo` is used to query the cluster information when the system is running on a cluster.
* `Router` is used to route the data to the destination worker so that it can be properly processed, with `PartitionInfo` and `ClusterInfo` as input.

## Related issue number

Fixes #2753


interactive_engine/executor/ir/graph_proxy/src/apis/router.rs
```diff
@@ -16,10 +16,12 @@
...
-    /// Get the local partition list that is going to be processed by current server.
-    fn get_local_partitions(&self) -> GraphProxyResult<Vec<u32>>;
+/// A router used to route the data to the destination worker.
+/// Specifically, given the graph element id to query,
+/// it firstly route the data to the server (aka. process) which is able to access the queried data from graph storage,
+/// and then pick a worker (aka. thread) in that server to do the query.
+pub trait Router: Send + Sync + 'static {
+    /// Given the element id and job_workers (number of workers per server),
+    /// return the worker id that is going to do the query.
+    fn route(&self, id: &ID, job_workers: usize) -> GraphProxyResult<u64>;
```
longbinlai, line 26: I have the following suggestions for this api:

- Design a structure, such like `ClusterInfo` here, to include how the whole cluster is managed, including how many servers (we actually need this, currently, we simply leverage a pegasus static function to obtain this information, which couples implementation with this api) in the cluster and how many workers in a server. The ClusterInfo can be initiated while starting the server, and then pass into this function as a reference. 
- Make the item of `id` as an abstraction of `RouteData`, where `RouteData` implements the function `get_route_id()`. Then we probably can directly implement `Vertex` and `Edge` (or even `ID`) as `RouteData`. `ID` is a little bit ambiguous given we actually want to route any data, not just `ID`.
-  Make `u64` a type of `WorkerId`.  
- I think Graph partition information should also be aware to the `Router`, no?
- Comment this structure as:
```
While processing 
```
A `Router` is used to route the data to the destination worker so that it can be properly processed, especially
when the underlying data has been partitioned across the cluster. Given the partition information as well as how our cluster is managed (by `ClusterInfo`) and co-located with the graph data, we can implement the corresponding `route` function to guide the system to transfer the data to a proper destination worker. 

For example, suppose our computer server contains 10 servers, each further forking 10 workers for processing queries. In addition, the graph is partitioned into these 10 servers by the following strategy: vertex of give ID is placed in the server with id i (0 to 9) i given ID % 10 == i, the vertex's adjacent edges are also placed with the vertex. Then the router can decide which worker should process the vertex of ID 25534 as follows:
- It first do `25534 % 10 == 4`, which means it must be routed to the 4-th server.
- Any worker in the 4-th server can process the vertex. Thus it randomly picks a worker, saying 5-th worker, which has ID 4 * 10 + 5 = 45.
- Then 45-th worker will be returned for routing this vertex.
```

File: interactive_engine/executor/ir/graph_proxy/src/apis/partitioner.rs [hunk 2]
```diff
@@ -16,13 +16,17 @@
...
-    fn get_worker_partitions(
-        &self, job_workers: usize, worker_id: u32,
-    ) -> GraphProxyResult<Option<Vec<u64>>>;
+pub type PartitionId = u32;
+pub type ServerId = u32;
+
+/// A `PartitionInfo` is used to query the partition information when the data has been partitioned.
+pub trait PartitionInfo: Send + Sync + 'static {
+    /// Given the data, return the id of the partition that holds the data.
+    fn get_partition_id(&self, data: &ID) -> GraphProxyResult<PartitionId>;
```
longbinlai, line 25: ID for data is not very nice. We may want to define `trait PartitionedData`, even we may not need any function for them.
```
pub trait PartitionedData {  }

impl PartitionedData for Vertex {  }
impl PartitionedData for Edge {  }

pub trait PartitionInfo: Send + Sync + 'static {
    fn get_partition_id<D: PartitionedData>(&self, data: &Data) -> GraphProxyResult<PartitionId>;
}

```

interactive_engine/executor/ir/runtime/src/router.rs
```diff
@@ -0,0 +1,59 @@
...
+/// For example, suppose our computer server contains 10 servers, each further forking 10 workers for processing queries.
+/// In addition, the graph is partitioned into these 10 servers by the following strategy:
+/// vertex of give ID is placed in the server with id i (0 to 9) i given ID % 10 == i, the vertex's adjacent edges are also placed with the vertex.
+/// Then the router can decide which worker should process the vertex of ID 25534 as follows:
+/// - It first do `25534 % 10 == 4`, which means it must be routed to the 4-th server.
+/// - Any worker in the 4-th server can process the vertex. Thus it randomly picks a worker, saying 5-th worker, which has ID 4 * 10 + 5 = 45.
+/// - Then 45-th worker will be returned for routing this vertex.
+pub trait Router: Send + Sync + 'static {
+    /// a route function that given the data, return the worker id that is going to do the query.
+    fn route(&self, data: &ID, job_workers: usize) -> GraphProxyResult<WorkerId>;
```
longbinlai, line 38: Similarly, `ID` for data, not very well-designed. Consider:
```
pub trait RouteData {
    /// Comment this
    fn get_route_key(&self) -> ID; 
}

impl RouteData for Vertex { ... }
impl RouteData for Edge { ... }
```

interactive_engine/executor/ir/runtime/src/router.rs
```diff
@@ -0,0 +1,59 @@
...
+/// we can implement the corresponding `route` function to guide the system to transfer the data to a proper destination worker.
+///
+/// For example, suppose our computer server contains 10 servers, each further forking 10 workers for processing queries.
+/// In addition, the graph is partitioned into these 10 servers by the following strategy:
+/// vertex of give ID is placed in the server with id i (0 to 9) i given ID % 10 == i, the vertex's adjacent edges are also placed with the vertex.
+/// Then the router can decide which worker should process the vertex of ID 25534 as follows:
+/// - It first do `25534 % 10 == 4`, which means it must be routed to the 4-th server.
+/// - Any worker in the 4-th server can process the vertex. Thus it randomly picks a worker, saying 5-th worker, which has ID 4 * 10 + 5 = 45.
+/// - Then 45-th worker will be returned for routing this vertex.
+pub trait Router: Send + Sync + 'static {
```
longbinlai, line 36:
```
pub trait Router {
    // Make PartitionInfo and ClusterInfo as type of Router to tell the implementation that Router needs these two information 
    type P: PartitionInfo; 
    type C: ClusterInfo; 
}
```

interactive_engine/executor/ir/runtime/src/router.rs
```diff
@@ -0,0 +1,59 @@
...
+    /// a route function that given the data, return the worker id that is going to do the query.
+    fn route(&self, data: &ID, job_workers: usize) -> GraphProxyResult<WorkerId>;
+}
+
+pub struct DistributedDataRouter<P: PartitionInfo, C: ClusterInfo> {
```
longbinlai, line 41: I would call this `DefaultRouter`

interactive_engine/executor/ir/runtime/src/router.rs
```diff
@@ -0,0 +1,59 @@
...
+}
+
+impl<P: PartitionInfo, C: ClusterInfo> DistributedDataRouter<P, C> {
+    pub fn new(partition_info: Arc<P>, cluster_info: Arc<C>) -> Self {
+        DistributedDataRouter { partition_info, cluster_info }
+    }
+}
+
+impl<P: PartitionInfo, C: ClusterInfo> Router for DistributedDataRouter<P, C> {
+    fn route(&self, data: &i64, job_workers: usize) -> GraphProxyResult<u64> {
```
longbinlai, line 53: Move the above comment:
```
/// For example, suppose our computer server contains 10 servers, each further forking 10 workers for processing queries.
/// In addition, the graph is partitioned into these 10 servers by the following strategy:
/// vertex of give ID is placed in the server with id i (0 to 9) i given ID % 10 == i, the vertex's adjacent edges are also placed with the vertex.
/// Then the router can decide which worker should process the vertex of ID 25534 as follows:
/// - It first do `25534 % 10 == 4`, which means it must be routed to the 4-th server.
/// - Any worker in the 4-th server can process the vertex. Thus it randomly picks a worker, saying 5-th worker, which has ID 4 * 10 + 5 = 45.
/// - Then 45-th worker will be returned for routing this vertex.
```

to here.

interactive_engine/executor/ir/runtime/src/router.rs
BingqingLyu: Yes, `RouteData` is a better definition. However, if we introduce `RouteData` in `Router`,  then `Router` cannot be made into an object. 
e.g., this would caused a compiling error:
```
#[derive(Clone)]
struct FnGenerator {
    router: Arc<dyn Router>,
}
```
Error:
```
error[E0038]: the trait `Router` cannot be made into an object
  --> runtime/src/assembly.rs:69:17
   |
69 |     router: Arc<dyn Router>,
   |                 ^^^^^^^^^^ `Router` cannot be made into an object
   |
note: for a trait to be "object safe" it needs to allow building a vtable to allow the call to be resolvable dynamically; for more information visit <https://doc.rust-lang.org/reference/items/traits.html#object-safety>
`` `

longbinlai (approved):
LGTM

PR #2773, opened 2023-05-30
[GIE Compiler] Fix query params of getV base in path expand operator

shirly121:
## What do these changes do?
In the cypher statement `(a:person)-[b:knows]-(c:person)` that represents a path expansion, the logical plan includes two `getV` operators. One is `getV` base inner the `path_expand` b, another is the `endV` operator following the b. It's important to note that these operators have different query parameters.

The `getV` base can retrieve vertices of any type that are adjacent to the type specified in the `path_expand` operation. **This pr specifically addresses this requirement**. The `endV` operator is constrained by the type specified in the given query.

## Related issue number

Fixes


longbinlai (approved):
LGTM

PR #2784, opened 2023-05-31
[Docs/GIE] add tinkerpop ecosystem

longbinlai:

Add back Tinkerpop's ecosystem from the old doc and re-org the docs a little bit.


BingqingLyu (approved):
LGTM.

PR #2793, opened 2023-06-02
Fixes the graphdef value when constructing graphs from vineyard_id.

### LINKED ISSUE #2792: [BUG] Reloaded graph from vineyard_id has no vineyard id
**Describe the bug**

Create a graph and reload it with its' vineyard id
```python
sess = graphscope.session()
graph = load_modern_graph(sess, "/testingdata/modern_graph")
new_graph = sess.g(vineyard.ObjectID(graph.vineyard_id))
```
Get the vineyard id and the result is 0
```
new_graph.vineyard_id
0
```

**Expected behavior**
The new graph's vineyard id should not be `0` and could be handled by the engines.

sighingnow:
## Related issue number

Fixes #2792


dashanji:
Thanks. I will try it.

dashanji (approved):

PR #2842, opened 2023-06-08
[GIE Doc] Add examples to collect gremlin results in a streaming way

shirly121:
## What do these changes do?
The official python gremlin client lacks guidelines about how to use streaming apis, here give a java example instead. The python example will be added later.

## Related issue number

Fixes


longbinlai (approved):
LGTM

Issue #2726, opened 2023-05-23
auto-generate docs for Flex


yecol:
Currently the reference doc files for Flex is pre-built and hosted in the code-repo, 
with path: `/flex/docs`

Please integrate the generate cmd to `make docs`, following the manner we used for the project:
- auto-gen the docs;
- please note that the generated files should be in `gh-pages` branch, rather than `main`;
- maybe output the generated files to a subfolder `Flex` under section `REFERENCES`.

ds-ssj:
In this issue, do we need to generate `html` files directly under REFERENCES (similar to the GAE JAVA API) or do we need to re-describe them in the style of `rst` files (similar to the GAE API and Python API) and then generate `html` files?

PR #2851, opened 2023-06-09
auto-gen docs of flex

ds-ssj:
## What do these changes do?

Auto-gen the docs of `flex` and remove the pre-built html files. Due to the reduction of dependencies at build time (e.g. using cmake for flex), a new Doxygen file was created, namely `Doxygen-flex`.

## Related issue number
#2726


yecol (approved):
Cool! Preventing our main coding lang from being `HTML`... 😭

cc @luoxiaojian: Please be aware these files are moving.

Issue #2824, opened 2023-06-06
[GAE/tests] Add test cases for FLASH algorithms in GAE


lixueclaire:
**Is your feature request related to a problem? Please describe.**
Currently, we provide tens of [built-in algorithms with the FLASH model](https://github.com/alibaba/GraphScope/tree/main/python/graphscope/analytical/app/flash) in GAE, but only three of them are added in the unit test. Thus we can add some new algorithms for testing.

**Describe the solution you'd like**
Add tests in this [test file](https://github.com/alibaba/GraphScope/blob/main/python/graphscope/tests/unittest/test_flash.py) for algorithms of: [bfs](https://github.com/alibaba/GraphScope/blob/main/python/graphscope/analytical/app/flash/traversal.py#LL59C1-L59C1) and [cc](https://github.com/alibaba/GraphScope/blob/main/python/graphscope/analytical/app/flash/connectivity.py#L64).
Please refer to the [documentation](https://graphscope.io/docs/latest/analytical_engine/tutorial_dev_algo_cpp_flash) about how to run FLASH algorithms.

PR #2869, opened 2023-06-13
[GAE/tests] Add GAE tests for FLASH algorithms of bfs and cc

longbinlai:

Fix issue #2824


lixueclaire (commented):
You could use `./gs format python` to format the code

lixueclaire (approved):
LGTM～

PR #2885, opened 2023-06-14
Allow pass params to gie instance

siyuan0322:
Allow passing a dict to customize the configuration of GIE instance.

```python
interactive = graphscope.gremlin(graph, params={'pegasus.timeout': 600000})
```

The key-value pair would be append to the `vineyard.frontend.properties` and `vineyard.executor.properties`.


longbinlai:
Can we call the name "query.timeout" instead of "pegasus.timeout" ?

siyuan0322:
> Can we call the name "query.timeout" instead of "pegasus.timeout" ?

So, all that is needed is to replace the text 'pegasus.timeout' with 'query.timeout' throughout the code base, or is more work required from @shirly121 ?

siyuan0322:
Merged, we could leave the renaming to the unify the timeout PR.

PR #2894, opened 2023-06-15
[GAE-Java] Relax version enforce for some mvn dependencies

zhanglei1949:
When upgrading `guava` for `GAE Java SDK`, there seems to be some problem with `maven-enforcer-plugin`. Relax this version control.

### DIFF
--- a/analytical_engine/java/grape-rdd-reader/pom.xml
+++ b/analytical_engine/java/grape-rdd-reader/pom.xml
@@ -83,7 +83,6 @@
         <dependency>
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
-            <version>31.0.1-android</version>
         </dependency>
         <dependency>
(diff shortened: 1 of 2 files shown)


analytical_engine/java/grape-rdd-reader/pom.xml
```diff
@@ -83,7 +83,6 @@
         <dependency>
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
-            <version>31.0.1-android</version>
         </dependency>
```
yecol, line 86: Are you missing a line here? 
`<version>${guava.version}</version>`

zhanglei1949: The version control of all dependencies in child module can be uplift to parent module, so we don't need to specify the version here, let me add one more commit.

sighingnow (approved):

PR #2906, opened 2023-06-15
[Bug Fix] Fix bugs of time counter in gremlin service

### LINKED ISSUE #2905: [BUG] Inaccurate Execution Time Calculation in Gremlin Service's Time Counter
**Describe the bug**
The Time Counter tool in the Gremlin Service is responsible for calculating the execution time per query. However, due to recent changes introduced in the PR https://github.com/alibaba/GraphScope/pull/2818, the behavior of the Time Counter has become inaccurate.

Previously, the compiler worker thread would wait for the response from the runtime and calculate the execution time, which was reasonable. However, after the mentioned PR was merged into the main branch, the worker thread no longer waits for the runtime response and sends the physical plan directly (grpc callback funtion will help to handle response). As a result, the execution time calculated by the Time Counter only reflects the compilation time, without accounting for the actual execution time by the runtime. This behavior is undesirable and needs to be addressed.

**To Reproduce**
1. Execute a query using the Gremlin Service.
2. Observe the reported execution time provided by the Time Counter.

**Screenshots**
image

**Environment (please complete the following information):**
 - GraphScope version: [main]
 - OS: [e.g. MacOS, Linux]

shirly121:
## What do these changes do?
as titled

## Related issue number

Fixes #2905


shirly121:
fix according to review suggestions, see https://github.com/alibaba/GraphScope/pull/2906/commits/76fda65dc6a88f97ecd7d1f82db2eefc818026d9: 
1. rename `locked` to `isContextWritable`
2. add QueryLogger to print log info per query

longbinlai (approved):
LGTM

PR #2915, opened 2023-06-19
[Coordinator] Fix issue #2734: unifying timeout parameter in coordinators

### LINKED ISSUE #2734: Unify `timeout` parameters within coordinator
The Coordinator contains many operations that do not return immediately, but wait for a period of time to succeed or return an error after a timeout.

These operations include but are not limited to: 
- creating Coordinator services (PODs)
- creating GAE services (PODs)
- creating GIE services (PODs)
- creating GLE services (PODs)
- stopping (deleting) the above services (PODs)
- and querying results from GIE. 
 
The timeout time in the Coordinator is quite fragmented, and each operation does not use a unified timeout parameter to control, which also makes it difficult to adjust the timeout time of each operation from the client. 

It is expected to use one single timeout parameter to control all the waiting timeouts in the Coordinator.

TaoLbr1993:
## What do these changes do?

1. Add timeout parameters for coordinations such that they can be adjusted via arguments;
2. Add retry time parameters for coordinations such that they can be adjusted via arguments;
3. Refine the logic of testing timeout to avoid the unwanted error if timeout is less than retry time;
4. Correct a typo in a function name.

## Related issue number

Fixes #2734

### DIFF
--- a/coordinator/gscoordinator/coordinator.py
+++ b/coordinator/gscoordinator/coordinator.py
@@ -387,7 +387,9 @@ class CoordinatorServiceServicer(
     def FetchLogs(self, request, context):
         while self._streaming_logs:
             try:
-                info_message, error_message = self._pipe_merged.poll(timeout=self._poll_timeout_seconds)
+                info_message, error_message = self._pipe_merged.poll(
+                    timeout=self._poll_timeout_seconds
+                )
             except queue.Empty:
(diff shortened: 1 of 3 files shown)


yecol:
hi @TaoLbr1993 , to pass the configured checks, you need to sign CLA.

siyuan0322 (approved):

PR #2959, opened 2023-07-03
Enable perfect hash support in GAE

sighingnow:


python/graphscope/framework/graph.py
```diff
@@ -265,7 +268,8 @@ def __init__(
                 Defaults to global.
             compact_edges (bool, optional): Compact edges (CSR) using varint and delta encoding. Defaults to False.
                 Note that compact edges helps to half the memory usage of edges in graph data structure, but may cause
-                at most 10%~20% performance degeneration in some algorithms.
+                at most 10%~20% performance degeneration in some algorithms. Defaults to False.
+            use_perfect_hash (bool, optional): Use perfect hash in vertex map to optimize the memory usage. Defaults to False.
```
siyuan0322, line 272: Add a comment saying it only works for oid type int64_t? It I understand correctly.

siyuan0322: Or just check the oid and log error.

siyuan0322 (dismissed):

python/graphscope/framework/graph.py
sighingnow: Works for string oid as well.

PR #3013, opened 2023-07-17
[BugFix] Fix bug when process pk scan of multiple labels

### LINKED ISSUE #3008: [GIE/BUG] Unexpected `IndexScan` with multiple labels
**Describe the bug**
The `g.V().hasLabel("person").has("id", 1)` should be optimized as `IndexScan`, which leverages the primary key index of the vertex as "id" is the primary key of "person" vertex. However, if user specifies multiple labels, such as
```
g.V().hasLabel("person", "country").has("id", 1)
```
GIE compilers returns "user defined error: Empty/Multiple labels in `IndexScan`".

**To Reproduce**
Start GIE service, and type in queries with Empty/Multiple labels followed by primary key predicate, as:
```
g.V().hasLabel("person", "country").has("id", 1)
```
or,
```
g.V().hasLabel().has("id", 1)
```

**Expected behavior**
The query should return as expected, either allow `IndexScan` to support multiple labels (if they both have the same primary key), or disable `IndexScan` for such case. 

**Screenshot**
image

BingqingLyu:
Committed-by: bingqing.lbq from Dev container

## What do these changes do?

As titled.

## Related issue number

Fixes #3008

### DIFF
--- a/interactive_engine/executor/ir/core/src/plan/logical.rs
+++ b/interactive_engine/executor/ir/core/src/plan/logical.rs
@@ -703,6 +703,9 @@ fn triplet_to_index_predicate(
     if meta.schema.is_none() {
         return Ok(None);
     }
+    if tables.is_empty() {
+        return Ok(None);
+    }
     let schema = meta.schema.as_ref().unwrap();
     let mut key = None;
     let mut is_eq = false;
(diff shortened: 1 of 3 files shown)


longbinlai (approved):
LGTM

interactive_engine/executor/ir/core/src/plan/logical.rs
```diff
@@ -1920,6 +1928,52 @@ mod test {
         );
     }
 
+    // e.g., g.V().hasLabel("person", "software").has("name", "John")
+    #[test]
+    fn scan_multi_labels_pred_to_idx_pred() {
```
longbinlai, line 1933: Lack a query case with result verification.

sighingnow (approved):

interactive_engine/executor/ir/core/src/plan/logical.rs
BingqingLyu: Add a test case in test_min.py, which runs queries on vineyard that supports primary key.

longbinlai (approved):

Issue #3014, opened 2023-07-17
Unify all proto files.


siyuan0322:
We have many components that contains their own protos scattered across the repo, it should be unified into one directory to simplify management.

- [ ] Python and Coordinator
- [ ] GAE
- [ ] GIE Compiler
- [ ] GIE Executor and Pegasus

PR #3023, opened 2023-07-18
unify gae, python and groot proto into proto directory

siyuan0322:
## Related issue number

Related to #3014


longbinlai:
Can we arrange the proto dir as the flex dir? with a stores dir for all graph store in graphscope, with an engines dir for all engines in graphscope?

coordinator/gscoordinator/kubernetes_launcher.py
```diff
@@ -121,6 +121,8 @@ def __init__(
         self._apps_api = kube_client.AppsV1Api(self._api_client)
         self._resource_object = ResourceManager(self._api_client)
 
+        self._serving = False
```
sighingnow, line 124: Already been fixed in https://github.com/alibaba/GraphScope/pull/3019

sighingnow (approved):

PR #3010, opened 2023-07-14
Implement grin interface for flex

liulx20:
## What do these changes do?

Implement grin interface for flex.

## Related issue number
.

PR #3056, opened 2023-07-26
fix: Fixes error in docs generation by removing the submodule

sighingnow:
The submodule was first introduced in #3010 and it should be removed before `git add -A` to gh-pages.

Fixes https://github.com/alibaba/GraphScope/actions/runs/5665819421/job/15351324239


yecol:
Thanks for completing the `pr-check.yml` !

PR #3063, opened 2023-07-28
feat(analytical): support use int32_t as vid_t to save memory for small graphs

sighingnow:


sighingnow:
Hi @dzhiwei, @wuyueandrew,

I would like you folks take a look on that pull request, which would be helpful for saving memory on small (less than 100M vertices) graphs, especially graphs don't have many properties.

siyuan0322 (approved):

Issue #3064, opened 2023-07-28
Include `GRAPE-jdk` in graphscope wheel package.


zhanglei1949:
Currently, `GRAPE-jdk` is not packed into GraphScope wheel package.

PR #3065, opened 2023-07-28
refactor(analytical): Remove `nativeLibLoader` in GRAPE-jdk

zhanglei1949:
Committed-by: zhanglei1949 from Dev container

Currently we pack the `libgrape-jni.so/dylib` into `grape-runtime-shaded-{}.jar` and  use `NativeLibLoader` to load the jni lib at runtime. But If we want to package `GRAPE-jdk` into GraphScope wheel package, we need to user `auditwheel` to reorg the dynamic linking and RPATH staffs.

So maybe we should abandon current approach, just copy `libgrape-jni` to /opt/graphscope/lib, such that is can be found by auditwheel.

## Related issue number

Related Issue: #3064

### DIFF
--- a/analytical_engine/java/grape-giraph/pom.xml
+++ b/analytical_engine/java/grape-giraph/pom.xml
@@ -78,10 +78,6 @@
       <artifactId>giraph-core</artifactId>
       <version>1.3.0-hadoop2</version>
     </dependency>
-    <!-- <dependency>
-      <groupId>org.scijava</groupId>
-      <artifactId>native-lib-loader</artifactId>
-    </dependency> -->
   </dependencies>
   <build>
     <plugins>
--- a/analytical_engine/java/grape-giraph/pom.xml
+++ b/analytical_engine/java/grape-giraph/pom.xml
@@ -78,10 +78,10 @@
       <artifactId>giraph-core</artifactId>
       <version>1.3.0-hadoop2</version>
     </dependency>
-    <dependency>
+    <!-- <dependency>
       <groupId>org.scijava</groupId>
       <artifactId>native-lib-loader</artifactId>
-    </dependency>
+    </dependency> -->
   </dependencies>
   <build>
     <plugins>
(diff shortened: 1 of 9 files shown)


analytical_engine/java/grape-giraph/pom.xml
```diff
@@ -78,10 +78,10 @@
       <artifactId>giraph-core</artifactId>
       <version>1.3.0-hadoop2</version>
     </dependency>
-    <dependency>
+    <!-- <dependency>
       <groupId>org.scijava</groupId>
       <artifactId>native-lib-loader</artifactId>
-    </dependency>
+    </dependency> -->
```
sighingnow, line 84: Remove if not needed. Just commenting it would leave a lot of confusion for other developers.

sighingnow (commented):
If I understand it correctly, the grape-jdk would be distributed both along with the wheel package and standalone as a jar from maven (or from mvn install). So, it would be better to use something like trying system-wide `LoadLibrary` first, and if failed, fallback to the `nativeLibLoader`.

analytical_engine/java/grape-giraph/pom.xml
zhanglei1949: Yep, will be removed.

lidongze0629 (approved):

zhanglei1949:
> If I understand it correctly, the grape-jdk would be distributed both along with the wheel package and standalone as a jar from maven (or from mvn install). So, it would be better to use something like trying system-wide `LoadLibrary` first, and if failed, fallback to the `nativeLibLoader`.

Just saying, if we are determined to install `libgrape-jni.so` to `${GRAPHSCOPE_HOME}/lib`, using nativeLibLoader seems dummy(cause it try to find .so in jar). And since `grape-runtime.jar` is not distributed via maven central repo, and we don't need to take care about backward compatibility. So, If user use the correct version of `grape-jdk`, the `Library Loading` will work as expected.

sighingnow (approved):

sighingnow (approved):

lidongze0629 (approved):

PR #3066, opened 2023-08-01
fix(analytical): fixes wrong result when flatten from projected arrow fragment

hanjf12:
fix error: wrong result when flatten from projected data

## Related issue number

Fixes


sighingnow (dismissed):

sighingnow:
@hanjf12 Could you please address the format issue by following the guidance in the message? https://github.com/alibaba/GraphScope/actions/runs/5722335994/job/15506017197?pr=3066

Thanks!

Issue #3058, opened 2023-07-26
[BUG] Combinations of Physical Operators Causes Issues with Pegasus


shirly121:
**To Reproduce**
1. launch graph instance locally by running:
```
cd interactive_engine/executor/ir
RUST_LOG=INFO ./target/release/start_rpc_server --config ./integrated/config

cd interactive/compiler
make run
```
2. Submit the following gremlin query:
```
g.V().order().by('id', asc).limit(1).out().count()
```
<img width="872" alt="image" src="https://github.com/alibaba/GraphScope/assets/22363306/6be3d9aa-8648-41e1-8121-36cd0bcc2d99">

PR #3067, opened 2023-08-01
fix(interactive): fix a bug of count operator

lnfjpt:
## What do these changes do?
Fix communication bug in `count` operator. Add corresponding unit test.

## Related issue number

Fixes
https://github.com/alibaba/GraphScope/issues/3058

### DIFF
--- a/interactive_engine/executor/engine/pegasus/pegasus/src/progress.rs
+++ b/interactive_engine/executor/engine/pegasus/pegasus/src/progress.rs
@@ -268,11 +268,21 @@ impl EndOfScope {
     }
 
     pub(crate) fn contains_source(&self, src: u32) -> bool {
-        (self.tag.is_root() && src == 0)
-            || (!self.tag.is_root()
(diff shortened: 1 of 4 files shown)


interactive_engine/executor/engine/pegasus/pegasus/src/progress.rs
```diff
@@ -269,7 +269,10 @@ impl EndOfScope {
 
     pub(crate) fn contains_source(&self, src: u32) -> bool {
         (self.tag.is_root() && src == 0)
-            || (!self.tag.is_root() && self.peers.value() > 0 && self.peers.contains_source(src))
+            || (!self.tag.is_root()
```
longbinlai, line 272: Consider using `if .. else` to simplify the condition check, like
```
if self.tag.is_root() {
    src == 0  
} else {
    // index: get the worker for handling the current subtask (add a function) 
    if src != index {
       (self.peers.value() > 0 && self.peers.contains_source(src)))
    } else {
        true 
    }
}
```

In addition, please properly comment this function to explain the given cases.

PR #3088, opened 2023-08-08
feat(interactive): Support Optional Match and Where Not(subquery) in Compiler

shirly121:
## What do these changes do?
1. support optional match as left outer join 
2. support where(not(subquery)) which is converted to anti join by `NotExistToAntiJoinRule`
3. support `FilterIntoJoinRule` to push down filters into join (using calcite standard implementation)

## Related issue number

Fixes


interactive_engine/compiler/conf/ir.compiler.properties
```diff
@@ -16,7 +16,7 @@ pegasus.hosts: localhost:1234
 # graph.schema
 graph.schema: ../executor/ir/core/resource/modern_schema.json
 graph.store: exp
-graph.planner: {"isOn":true,"opt":"RBO","rules":["FilterMatchRule"]}
+graph.planner: {"isOn":true,"opt":"RBO","rules":["FilterIntoJoinRule", "FilterMatchRule", "NotExistToAntiJoinRule"]}
```
longbinlai, line 19: `NotExist` looks like a gremlin semantics. How about `NotMatch`?

shirly121: fix in https://github.com/alibaba/GraphScope/pull/3088/commits/03fed995ff05e005d1b37ca44ad19046dfa5ee24

longbinlai (approved):

Issue #3100, opened 2023-08-11
[Bug/GIE]  `gs.interactive()` in python SDK reports error


longbinlai:
**Describe the bug**
Following the [instruction](https://graphscope.io/docs/latest/interactive_engine/getting_started), 
while calling `g = gs.interactive(graph)`, there reports error of `'module' object is not callable`. 

**Screenshots**
![Screenshot 2023-08-11 at 16 32 45](https://github.com/alibaba/GraphScope/assets/7007589/ce540199-4070-417d-9fcf-ab6be3453c09)

The above error turns out to be an inconsistency between the latest release and the doc. 
The doc has included the changes in pr #2952, which however, has not been encompassed in the latest release of graphscope (0.23.0). Using preview version (`pip3 install graphscope --pre`) may resolve the ''`module not callable`'' issue, but new issue prompts:
![Screenshot 2023-08-14 at 11 31 18](https://github.com/alibaba/GraphScope/assets/7007589/c1e04d41-1c38-4666-8364-c731ad3280eb)

![Screenshot 2023-08-14 at 11 32 31](https://github.com/alibaba/GraphScope/assets/7007589/3d6fea2c-5104-48f3-bc0f-e6bc8d3e3585)

 

Furthermore, the result printing is wrong in the doc as well, should be `print(q1.all())` instead of `print(q1.all().result())` 

![Screenshot 2023-08-11 at 16 34 44](https://github.com/alibaba/GraphScope/assets/7007589/7eabc9ec-36ca-4cf4-998c-006ce10ea61c)

**Environment (please complete the following information):**
 - GraphScope version: 0.23.0
 - OS: MaxOS
 - Version 13.3.1

shirly121:
I successfully go through the process in the Dev Container environment.
```
>>> q1 = g.execute('g.V().count()')
>>> print(q1.all().result())
[6]
>>> q2 = g.execute('g.V().hasLabel(\'person\')')
>>> print(q2.all().result())
[v[2], v[3], v[0], v[1]]
>>> from neo4j import GraphDatabase, RoutingControl
>>> q3 = g.execute("MATCH (n:person) RETURN count(n)", lang="cypher", routing_=RoutingControl.READ)
>>> print(q3.records[0][0])
4
>>> q3 = g.execute("MATCH (n) RETURN count(n)", lang="cypher", routing_=RoutingControl.READ)
>>> print(q3.records[0][0])
6
```

longbinlai:
[...]

PR #3131, opened 2023-08-21
fix(interactive): fix the docs of interactive engine

longbinlai:

Fix issue #3100


docs/interactive_engine/getting_started.md
```diff
@@ -6,11 +6,13 @@ This guide gives you a quick start to use GraphScope for graph interactive tasks
 
 GIE works as the core component of GraphScope. Therefore, you should install GraphScope at first to use GIE.
 
-GraphScope provides python interface which requires:
+GraphScope requires the following software versions:
 
-- Python3.7 ~ 3.11
+- Python 3.9 ~ 3.11
+- Java 11 (both Java 8 and 20 have compatibility issues)
```
sighingnow, line 12: Maybe it would be better to list the environment requirement for all engines in the same place? it would be easier to keep them consistent.

e.g., you'll find that the following page is saying we only require Java 8 for GIE.

https://graphscope.io/docs/deployment/install_on_local#prerequisites

longbinlai: Make sense. But shall we include an extra page for prerequisites. It 's actually required everywhere, and there are much more requirements than we've listed in https://graphscope.io/docs/deployment/install_on_local#prerequisites.

longbinlai: @sighingnow I have already tried, java 8 would not work for gie any more. LoL

sighingnow (dismissed):

shirly121 (approved):

Issue #3132, opened 2023-08-21
[BUG] Unexpected results of query `Apply` with `Iterate` in subtask in Pegasus


BingqingLyu:
**Describe the bug**

The following query (on LDBC dataset) output errors:

```
g.V().hasLabel('forum').as('forum')
 .select('forum').by(out('containerOf').in('1..3', 'replyOf').endV().has('id', 6047318147162))
```

The error info client gets is:
```
GremlinServerError: 500: Status{code=UNKNOWN, description=execution_error: system error: kind(IOError), caused 
by:IOError(kind=System(NotConnected)),   occurred at: 
/work/interactive_engine/executor/engine/pegasus/pegasus/src/data_plane/intra_process.rs:43, ;, cause=null}
```

and internal error info (in pegasus) is:
```
thread 'reactor 2' panicked at 'iteration of [13] not found;', /root/moran/new/GraphScope/interactive_engine/executor/engine/pegasus/pegasus/src/operator/iteration/switch.rs:271:21
```

**Additional context**

As an optional solution, we can modify the query like:
```
g.V().hasLabel('forum').as('forum').select('forum').
by(out('containerOf').union(__.in('replyOf'), __.in('replyOf').in('replyOf')).has('id', 6047318147162).as('message'))
```

PR #3133, opened 2023-08-21
fix(interactive): throw unsupported error when query `PathExpand()` within subtask

BingqingLyu:
## What do these changes do?

Currently, throw unsupported error when query `PathExpand()` (which is implemented by `Iterate`) within subtask in physical plan in GIE. 

Once Pegasus supports `Apply` with `Iterate` in subtask, this can be removed.

## Related issue number

#3132


lnfjpt (approved):
LGTM

longbinlai (approved):

PR #3145, opened 2023-08-25
refactor(interactive):  Parse Graph Schema from Yaml Configuration

shirly121:
## Related issue number

Fixes


shirly121:
@longbinlai infer file format from its content instead of extension: https://github.com/alibaba/GraphScope/pull/3145/commits/4d33e52d1ca22bba45f34b41a4bdbddbee8a66ac

longbinlai (approved):

interactive_engine/compiler/src/test/resources/schema/modern.yaml
zhanglei1949, line 1: Can we reuse interactive's default graph.yaml in [interactive dierectory](https://github.com/alibaba/GraphScope/blob/main/flex/interactive/data/ldbc/graph.yaml), rather than creating a new file only for testing?

zhanglei1949: It seems reasonable to keep them in src/test/resources.

zhanglei1949 (approved):

Issue #3074, opened 2023-08-03
Upgrade the build phase of image in CI with gsctl


lidongze0629:
**Is your feature request related to a problem? Please describe.**

See background in #2925. 

For this task, you may want to investigate all the build phases in the CI, and replace them using `gsctl make --xxx`, to make the related sub-commands in `gsctl` robust and confident.

Here are some tips:

- after checkout, the first command should be cd python/graphscope && pip install --editable . to make gsctl usable.
add the missing pieces, (options or subcommand) in gsctl, if any.
- Please note there is a CI env in github actions. Check it if you have specific codings for CI runtime.
- To make our life easier, you may leave build wheel related jobs untouched.

**Here are the usage:**
```
gsctl install-deps     Install dependencies on local machine.
gsctl make xxx         # Build GraphScope on local.
gsctl test                  # Run tests of graphscope
gsctl format            
gsctl dev                   # Open a develop environment with docker.
```

**Useful links**
-  [cli (to be deprecated)](https://github.com/GraphScope/cli)
- [python click](https://click.palletsprojects.com/en/8.1.x/)

lidongze0629:
![截屏2023-08-03 上午11 28 05](https://github.com/alibaba/GraphScope/assets/23491543/a6b62c2c-2d53-4d65-884c-d05324ec9fda)

lidongze0629:
pr 形式:
- 删除 gs 文件
- 重构 gsctl 文件，支持已有的 gs 支持的功能，保证 CI 通过

lidongze0629:
pr 形式:
- 删除 gs 文件
- 重构 gsctl 文件，支持已有的 gs 支持的功能，保证 CI 通过

FLYLX:
gs文件在GraphScope文件下的命令整理：
gs --version | -v
gs --help | -h
gs make-image -h
gs install-deps :
        gs install-deps -h 
        
        gs install-deps dev:        
                gs install-deps dev
                gs install-deps dev --cn
                gs install-deps dev --v6d-version v0.16.1
                gs install-deps dev --cn --for-analytical --no-v6d
                gs install-deps dev --v6d-version=$VINEYARD_VERSION -j $(nproc)
                gs install-deps dev --v6d-version=$VINEYARD_VERSION -j 2
                gs install-deps dev --for-analytical --v6d-version=$VINEYARD_VERSION -j 2


gs make:
         gs make all
         gs make install:
                 gs make install
[...]

PR #3149, opened 2023-08-28
refactor: Upgrade gsctl and modify files associated with gsctl

FLYLX:
## What do these changes do?

Added the test method to gsctl.
Modify files associated with gsctl.
## Related issue number

issue #3074

### DIFF
--- a/docs/development/dev_guide.md
+++ b/docs/development/dev_guide.md
@@ -37,10 +37,10 @@ To use the dev containers for GraphScope development, you can follow these steps
 To install all dependencies on your local, use the GraphScope command-line utility [gs](https://github.com/alibaba/GraphScope/blob/main/gs) with the subcommand `install-deps` like this
 
 ```bash
-./gs install-deps dev
+python3 gsctl.py install-deps dev
 
 # for more usage, try
-# ./gs install-deps -h
+# python3 gsctl.py install-deps -h
 ```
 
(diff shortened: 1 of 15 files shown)


docs/development/dev_guide.md
```diff
@@ -37,10 +37,10 @@ To use the dev containers for GraphScope development, you can follow these steps
 To install all dependencies on your local, use the GraphScope command-line utility [gs](https://github.com/alibaba/GraphScope/blob/main/gs) with the subcommand `install-deps` like this
 
 ```bash
-./gs install-deps dev
+gsctl install-deps dev
```
siyuan0322, line 40: By using `gsctl xxx` directly, you are assuming the `gsctl` could be found in the `PATH`. Either you should add a step to install `gsctl` to the system path, or you need to add the path to gsctl, for example: `./gsctl xxx`.

This applies to all other `gsctl` usages throughout the documents.

python/graphscope/gsctl/scripts/test_command.sh
```diff
@@ -76,7 +77,7 @@ function test_analytical-java {
 
 function test_interactive {
```
lidongze0629, line 78: Unify the indentation of functions in all scripts

lidongze0629 (approved):

PR #3160, opened 2023-08-31
fix(interactive): fixes some unsafe rust code that cannot be compiled with recent stable rust compilers

sighingnow:
## What do these changes do?

This pull request fixes some incompatible issues with recently stable Rust compilers (>=1.72.0).

- TypeId casting:
  
  The `TypeId` in rust becomes `i128` since https://github.com/rust-lang/rust/commit/9e5573a0d275c71dce59b715d981c6880d30703a, makes the original `std::mem::transmute` stop working.

- The ref to mut-ref `&mut *(t as *const T as *mut T)` has become undefined behaviour since Rust 1.70, and was rejected by recent nightly compilers.

- The `EdgeInfo::kinds` yields a anti-pattern design for Rust (a constant Vec is modified without any protection). This pull request adds a lock to the `kinds` vector, and refactors its iter implementation to make it safer (and, can be compiled).

### DIFF
--- a/interactive_engine/executor/assembly/groot/src/executor/gaia/gaia_library.rs
+++ b/interactive_engine/executor/assembly/groot/src/executor/gaia/gaia_library.rs
@@ -20,7 +20,6 @@ use std::sync::Arc;
 
 use groot_store::db::api::GraphConfigBuilder;
 use groot_store::db::common::bytes::util::parse_pb;
-use groot_store::db::common::unsafe_util::to_mut;
 use groot_store::db::graph::store::GraphStore;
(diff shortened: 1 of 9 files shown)


siyuan0322 (approved):

longbinlai:
Thanks @sighingnow. Will these issues also affect code of pegasus and ir-core? @BingqingLyu

sighingnow:
> Thanks @sighingnow. Will these issues also affect code of pegasus and ir-core? @BingqingLyu

Merged as CI turns green.

PR #3167, opened 2023-09-01
feat(interactive): Support `IS_NULL` and `IS_NOT_NULL` in Cypher Queries

shirly121:
## What do these changes do?
as titled.

## Related issue number

Fixes


interactive_engine/compiler/src/test/java/com/alibaba/graphscope/cypher/antlr4/WhereTest.java
```diff
@@ -173,4 +175,90 @@ public void where_6_test() {
                     + "], matchOpt=[INNER])",
                 after.explain().trim());
     }
+
+    // 'b is null' cannot be pushed down, for the conversion will change the semantics
+    @Test
+    public void where_7_test() {
+        RelNode multiMatch =
+                Utils.eval("Match (a) Optional Match (a)-[]->(b) Where b is null Return a").build();
```
longbinlai, line 183: how do we control whether `b is null` can or cannot be pushed down? how about
```cypher
Match (a) Optional Match (a)-[]->(b) Where b.age is null Return a
```
can b.age is null be pushed down?

shirly121: The official Calcite `FilterIntoJoinRule` determines whether filter conditions can be pushed down into a join. The condition `b.age is null` is actually equivalent to `b is not null and b.age is null`, and in such cases, it can be pushed down into the join.
However, currently, when the compiler interfaces with the Calcite type system, it assumes that types coming from storage cannot be null, so `b.age is null` is inferred to produce no results and generate the following logical plan:
```
// Match (a) Optional Match (a)-[]->(b) Where b.age is null Return a
GraphLogicalProject(a=[a], isAppend=[false])
  LogicalValues(tuples=[[]])
```
Here `LogicalValues` denotes a no result plan

longbinlai (approved):

PR #3168, opened 2023-09-01
feat(interactive): Support `[]` Operator in Cypher Queries

shirly121:
## What do these changes do?
1. support `variable list` operator which is denoted by `[a.name, a.age]` in cypher queries
2. introduce a new type for `GraphPath` to differentiate it from other ArrayTypes, as the `length()` function can only be applied to `GraphPath` types

## Related issue number

Fixes


longbinlai (approved):

interactive_engine/compiler/src/test/java/com/alibaba/graphscope/cypher/antlr4/WithTest.java
```diff
@@ -184,4 +184,22 @@ public void with_10_test() {
                     + "], matchOpt=[INNER])",
                 project.explain().trim());
     }
+
+    @Test
+    public void with_11_test() {
+        RelNode project =
+                Utils.eval("Match (a:person)-[]-(b:person) Return [a.name, b.age, 1]").build();
```
longbinlai, line 191: how about returning an empty array like `[ ]`

shirly121: > how about returning an empty array like `[ ]`

We have the same behavior as official Neo4j, both of which output an empty list.
```
@neo4j> Match (a) Return [];
+-----+
| $f0 |
+-----+
| []  |
| []  |
| []  |
| []  |
| []  |
| []  |
+-----+
```

longbinlai (approved):

Issue #3137, opened 2023-08-23
feat (interactive): Implement field trimmer


JialuGong:
**Is your feature request related to a problem? Please describe.**
Currently, GIE lacks the ability to trim unused fields. 

**Describe the solution you'd like**
Implement a calcite-like field trimmer for GIE. 
- [ ] Do field trimming by changing the current `Project RelNode`s  and `TableScan RelNode`s
- [ ] Add necessary `Project` between `RelNode`s  to trim unused fields

PR #3200, opened 2023-09-08
fix(interactive): Implement Field Trimmer Rule in Compiler

JialuGong:
## Related issue number
 #3137

Just a draft pr for pre-review


interactive_engine/compiler/src/main/java/com/alibaba/graphscope/common/ir/planner/GraphFieldTrimmer.java
```diff
@@ -0,0 +1,214 @@
...
+        return dispatchTrimFields(root, fields);
+    }
+
+
+    public RelNode trimFields(GraphLogicalProject project, ImmutableSet<GraphVariable> fieldsUsed) {
+        // TODO(huaiyu)
+        return project;
+    }
+
+    public RelNode trimFields(GraphLogicalAggregate aggregate, ImmutableSet<GraphVariable> fieldsUsed) {
```
shirly121, line 90: There's no need to implement a TrimField function for every operator, the difference only occurs when collecting the variables used by the operator. The rest of the process can be unified.

shirly121 (approved):

BingqingLyu (approved):

PR #3230, opened 2023-09-18
feat(learning): Integrate GLTorch into GraphScope

LiSu:
This PR integrates [GLTorch](https://github.com/alibaba/graphlearn-for-pytorch) into GraphScope, allowing training GNNs using GLTorch in the local mode of GraphScope.

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -149,7 +149,7 @@ class BuildGLTorchExt(torch.utils.cpp_extension.BuildExtension if torch else bui
     def _get_gcc_use_cxx_abi(self):
         if hasattr(self, "_gcc_use_cxx_abi"):
             return self._gcc_use_cxx_abi
-        build_dir = os.path.join(glt_root_path, 'cmake-build')
+        build_dir = os.path.join(glt_root_path, "cmake-build")
         os.makedirs(build_dir, exist_ok=True)
         output = subprocess.run(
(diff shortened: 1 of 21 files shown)


python/setup.py
```diff
@@ -111,12 +127,54 @@ def run(self):
...
+    def _get_gcc_use_cxx_abi(self):
+        if hasattr(self, "_gcc_use_cxx_abi"):
+            return self._gcc_use_cxx_abi
+        output = subprocess.run(
+            [shutil.which("cmake"), "."],
+            cwd=glt_root_path,
+            capture_output=True,
+            text=True,
+        )
+        import re
```
sighingnow, line 158: Maybe running cmake in a clean build directory would be better, i.e.,

```
build_dir = os.path.join(glt_root_path, 'cmake-build')
os.makedirs(build_dir, exist_ok=True)
output = subprocess.run([shutil.which('cmake'), '..'], cwd=build_dir, ....)
```

sighingnow (dismissed):
LGTM.

python/setup.py
LiSu: Fixed by adding a clean cmake build directory.

sighingnow (approved):
LGTM.

The failure in Python test is irrelevant with this PR.

PR #3240, opened 2023-09-20
fix(interactive): Not requiring eid in update/delete edges in groot

siyuan0322:
- Add API for writing vertices and edges in one batch
- Add eid to the Edge class in client
- Update an not existing edge will create a new edge.
- Update and Delete when eid is not given will try to search for it by src and dst vertex first.


sighingnow:
💪

Issue #2648, opened 2023-04-25
Enabling Natural Language Graph Queries in GraphScope with chatGPT


acezen:
**Is your feature request related to a problem? Please describe.**
Graph queries are a critical workload in GraphScope, currently submitted through query languages including Gremlin and Cypher. Although these graph query languages greatly reduce the user barrier compared to programming languages like Java or C++, GraphScope users still frequently struggle with query writing. This not only severely limits user graph querying capabilities, but also consumes valuable developer time in answering user questions. With the emergence of chatGPT, people have realized the incredible ability of large language models to handle natural language-related tasks including writing articles, answering questions, and even writing code. We wonder if we can leverage the ability of large language models to further reduce the barrier to using graph queries in GraphScope.

In this task, you will explore this question with us. Specifically, we expect that given the context of graph queries, such as query samples, graph schema information, and reference query implementations (based on Cypher), can we allow users to automatically generate Cypher queries expressed in natural language?

**Describe the solution you'd like**
- Explore the use of chatGPT to generate additional test samples (including manual labeling) based on existing query samples.
- Use the above test samples to explore the improvement of Cypher query generation accuracy through the prompt method.
- Add natural language querying capabilities to the existing Python interface in GraphScope.
- Write a paper or technical report detailing the technical implementation, performance evaluation, and future development direction of the method.

AugensternF:
wanna join🙋

Suchun-sv:
wanna join🙋

longbinlai:
> 

Hi, please sign in as candidate in: https://m.summer-ospp.ac.cn/org/prodetail/2313e0395

Undertone0809:
I wonder what's the next step for this feature.

PR #3271, opened 2023-09-28
feat: Enabling Natural Language Graph Queries in GraphScope with chatGPT #2648

Suchun-sv:
## What do these changes do?

[TL;DR] We add LLM support to enhance the basic graph queries, allowing users to interact with graphs more casually.

> Specifically, the detailed steps can be shown as:
1. Using the Langchain to send natural language sentences to LLM (chatGPT, chatGLM, and so on) with manually designed prompts.
2. Adding the schema of graphs to form more expected results from LLM.
3. Extract Cypher statements from the results.
4. Running Cypher on the Graphs through the Cypher interactive backend.
5. Return results from the backend.

> Our contribution can be summarized as：

_1. Well-designed Langchain pipeline as well as prompt templates  (langchain_cypher.py).
2. Exposed the query interface (query.py::query_to_cypher)._

## Related issue number

We are related to the issue #2648 

Fixes


longbinlai (approved):
LGTM

Issue #3277, opened 2023-10-09
feat(Interactive): Support Index predicate for dynamic params


zhanglei1949:
Currently IndexScan with primary key is not supported. i.e.
```cypher
MATCH(v : PERSON {id : $personId }) return v.firstName;
```

PR #3282, opened 2023-10-11
fix(interactive): Support Index predicate for dynamic params in IR-Core

BingqingLyu:
## What do these changes do?

As titled.

## Related issue number

#3277


interactive_engine/executor/ir/proto/algebra.proto
```diff
@@ -191,10 +191,16 @@ message Limit {
 // where the values referred by k1, k2, ... are indexed and hence the
 // predicate can be efficiently verified by leveraging the index.
 message IndexPredicate {
+  message PkValue {
```
longbinlai, line 194: Looks like `PKValue` is not quite necessary here, Simply define 
```
message Triplet {
    common.Property key = 1;
    one of value {
         common.Value const = 2;
         common.DynamicParam param = 3;
    }
}
```

BingqingLyu: Done

longbinlai (approved):
Can merge as long as cis passed

zhanglei1949 (approved):
LGTM

PR #3285, opened 2023-10-12
fix(interactive): Distinguish 'label equals' or 'unique key equals' from Filter Conditions

shirly121:
## What do these changes do?
1. distinguish 'label equals' or 'unique key equals' from filter conditions, 'label equals' is fused to decide the label type while 'unique key equals' is fused for index predicate.
2. convert index predicate from RelNode to ir core.
3. handle special cases when converting 'SEARCH' operator from RelNode to ir core. Specifically, if 'SEARCH' happens in index predicate or the operand of 'SEARCH' is a continuous range, it will be converted to multiple conditions connected by 'and' or 'or'. These conversions may not necessarily be reasonable, at least in order to make ir core supported.

## Related issue number

Fixes


longbinlai (approved):
LGTM.

longbinlai (approved):

PR #3297, opened 2023-10-18
refactor(coordinator): and gsctl to support FLEX interactive

lidongze0629:
## Related issue number

Fixes


coordinator/gscoordinator/servicer/graphscope_one/service.py
```diff
@@ -0,0 +1,625 @@
...
+from gscoordinator.utils import create_single_op_dag
+from gscoordinator.version import __version__
+
+__all__ = ["GraphScopeOneServiceServicer", "init_graphscope_one_service_servicer"]
+
+logger = logging.getLogger("graphscope")
+
+
+def catch_unknown_errors(response_on_error=None, using_yield=False):
+    """A catcher that catches all (unknown) exceptions in gRPC handlers to ensure
```
sighingnow, line 76: Put it to utils module.

python/graphscope/config.py
```diff
@@ -325,9 +322,17 @@ class SessionConfig:
 
 @dataclass
 class Config(Serializable):
+    # Solution under the FLEX architecture, choose from 'GraphScope One', 'Interactive' or 'GraphScope Insight'
+    solution: str = "GraphScope One"
```
sighingnow, line 326: Maybe using an enum would better, like

```python
from enum import Flag, STRICT, auto
class StrictFlag(Flag, boundary=STRICT):
    RED = auto()
    GREEN = auto()
    BLUE = auto()
```

python/graphscope/gsctl/commands/dev_command.py
```diff
@@ -0,0 +1,366 @@
...
+            "clean",
+            "all",
+        ],
+        case_sensitive=False,
+    ),
+    required=False,
+)
+@click.option(
+    "--repo-home",
+    envvar="REPO_HOME",
```
sighingnow, line 174: Would `GRAPHSCOPE_REPO` or `GRAPHSCOPE_REPOSITORY` be better?

python/graphscope/gsctl/config.py
```diff
@@ -0,0 +1,161 @@
...
+        # set
+        self._current_context = context.name
+        self._contexts[context.name] = context
+
+        # write
+        contexts = [v.to_dict() for _, v in self._contexts.items()]
+        with open(GS_CONFIG_DEFAULT_LOCATION, "w") as file:
+            yaml.dump(
+                {"contexts": contexts, "current-context": self._current_context}, file
+            )
```
sighingnow, line 89: I suggest leave a message like "Switched the context to ...." whenever you touching the global configuration file, e.g.,

```bash
$ kubectl config use-context default
Switched to context "default".
```

python/graphscope/gsctl/config.py
```diff
@@ -0,0 +1,161 @@
...
+"""Global configuration"""
+
+import os
+import random
+from string import ascii_letters
+
+import yaml
+
+GS_CONFIG_DEFAULT_LOCATION = os.environ.get(
+    "GSCONFIG", os.path.expanduser("~/.gs/config")
```
sighingnow, line 28: I remember that we use `~/.graphscope` rather than `~/.gs` before. Keep it consistent would be better.

sighingnow (approved):
LGTM.

python/graphscope/gsctl/config.py
lidongze0629: Yes, we do use ~/.graphscope instead of ~/.gs. It is better to maintain consistency.

python/graphscope/gsctl/config.py
lidongze0629: Thanks for your suggestion. I will leave related messages  in switching `context` func

python/graphscope/gsctl/commands/dev_command.py
lidongze0629: Change to `GRAPHSCOPE_REPO ` and `--graphscope-repo`

sighingnow (approved):

PR #3298, opened 2023-10-19
fix(interactive): Fix the problem of not returning properties together when returning a vertex/edge.

zhanglei1949:
For cypher query like `MATCH(v) return v`, previously, we only returned the label and internal ID. With this PR, we will also return the properties of the vertex.

- Add `graph` for `Sink`'s input paramters.
- EdgeSet constructors need edge property names provided name.


flex/engines/hqps_db/core/operator/sink.h
```diff
@@ -175,13 +176,80 @@ void template_set_tuple_value(results::Collection* collection,
...
+    value->set_i64(any.value.d.milli_second);
+    break;
+  case gs::PropertyType::kString:
+    value->mutable_str()->assign(any.value.s.data(), any.value.s.size());
+    break;
+  case gs::PropertyType::kDouble:
+    value->set_f64(any.value.db);
+    break;
+  default:
+    LOG(FATAL) << "unknown type";
```
longbinlai, line 199: Better with `unsupported` and log any.type as well. `Log(Warn)` instead of FATAL?

zhanglei1949: Fixed.

longbinlai (approved):

luoxiaojian (approved):

PR #3301, opened 2023-10-19
feat(interactive): Reuse compiled native lib for same physical plan

zhanglei1949:
Add a server-side query cache on hqps engine.
- Add `plan_id` field in `PhysicalPlan`.
- Maintain the mapping between plan_id between compiled dynamic lib path.
- Avoid unnecessary recompilation for same physical plan.
- Each shard in hqps engine will have exactly one actor handling codegen request, which is actually delegated to `hiactor::thread_resource_pool`. And the thread will not be blocked.


LiSu (commented):
The actor part LGTM.

luoxiaojian (approved):

longbinlai (approved):

PR #3304, opened 2023-10-22
fix(interactive): Update example movie graph and fix some bug

zhanglei1949:
Update the example movie graph's schema and raw data for interactive. 
Add some demo cypher queries.


BingqingLyu (approved):
LGTM.

shirly121 (approved):

PR #3307, opened 2023-10-23
fix(interactive): fix bug in release cstr pointer in compiler

BingqingLyu:
## What do these changes do?

As titled.

## Related issue number

Fixes


siyuan0322 (approved):

interactive_engine/compiler/src/main/java/com/alibaba/graphscope/common/jna/type/FfiString.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    public String getValue() {
+        return this.value;
+    }
+
+    private void release() {
+        Pointer cstr = getPointer();
+        if (cstr != null) {
+            logger.info("free ffi str pointed by {}", cstr);
```
siyuan0322, line 44: Will this be too verbose in practice?

longbinlai: Make it `debug` makes more sense.

shirly121: done

PR #3312, opened 2023-10-25
fix(interactive): Introduce `Degree Fusion` RBO Rules

shirly121:
## What do these changes do?
fuse the following situations to expandDegree + sum:
1. expand+getV+count
2. expand+count

## Related issue number

Fixes


interactive_engine/compiler/src/main/java/com/alibaba/graphscope/common/ir/planner/rules/DegreeFusionRule.java
```diff
@@ -0,0 +1,244 @@
+package com.alibaba.graphscope.common.ir.planner.rules;
```
longbinlai, line 1: Copyright?

longbinlai (commented):
Can we manage a doc page for all these optimization rules?

interactive_engine/compiler/src/main/java/com/alibaba/graphscope/common/ir/planner/rules/DegreeFusionRule.java
shirly121: done

shirly121:
> Can we manage a doc page for all these optimization rules?

done

PR #3321, opened 2023-10-31
fix(interactive): Make Pegasus Related Settings Configurable Per Query

### LINKED ISSUE #3318: Support configure pegasus worker num per query in GIE
Like how we could configure the timeout value per query: `g.with('evaluationTimeout', 500).xxx`, add support for configure pegasus worker num.

shirly121:
## What do these changes do?
make pegasus.XX settings configurable per query:
```
g.with(pegasus.worker.num, 2).V()
g.with(pegasus.batch.size, 64).V()
g.with(pegasus.output.capacity, 64).V()
```
the priority is : query > system env > system property > config file

## Related issue number

Fixes #3318

### DIFF
--- a/interactive_engine/compiler/src/main/java/com/alibaba/graphscope/gremlin/plugin/processor/IrStandardOpProcessor.java
+++ b/interactive_engine/compiler/src/main/java/com/alibaba/graphscope/gremlin/plugin/processor/IrStandardOpProcessor.java
@@ -43,20 +43,23 @@ import com.alibaba.graphscope.gremlin.plugin.script.AntlrGremlinScriptEngineFact
 import com.alibaba.graphscope.gremlin.plugin.strategy.ExpandFusionStepStrategy;
 import com.alibaba.graphscope.gremlin.plugin.strategy.RemoveUselessStepStrategy;
 import com.alibaba.graphscope.gremlin.plugin.strategy.ScanFusionStepStrategy;
+import com.alibaba.graphscope.gremlin.plugin.traversal.IrCustomizedTraversal;
+import com.alibaba.graphscope.gremlin.plugin.traversal.IrCustomizedTraversalSource;
 import com.alibaba.graphscope.gremlin.result.processor.AbstractResultProcessor;
 import com.alibaba.graphscope.gremlin.result.processor.GremlinResultProcessor;
 import com.alibaba.pegasus.RpcClient;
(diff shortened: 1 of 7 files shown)


interactive_engine/compiler/src/main/java/com/alibaba/graphscope/gremlin/plugin/processor/IrStandardOpProcessor.java
```diff
@@ -329,17 +337,36 @@ protected void processTraversal(
                 PegasusClient.JobConfig.newBuilder()
                         .setJobId(jobId)
                         .setJobName(jobName)
-                        .setWorkers(PegasusConfig.PEGASUS_WORKER_NUM.get(configs))
-                        .setBatchSize(PegasusConfig.PEGASUS_BATCH_SIZE.get(configs))
-                        .setMemoryLimit(PegasusConfig.PEGASUS_MEMORY_LIMIT.get(configs))
-                        .setBatchCapacity(PegasusConfig.PEGASUS_OUTPUT_CAPACITY.get(configs))
+                        .setWorkers(PegasusConfig.PEGASUS_WORKER_NUM.get(queryConfigs))
```
siyuan0322, line 340: Will the `configs` value be preserved?
Consider this scenario:
1. The default PegasusConfig.PEGASUS_WORKER_NUM = 4 if not specified
2. Ir Server is created with config pegasus.worker.num = 2
3. Specify pegasus.worker.num = 1 by `g.with(xxx,xx)`
The latter settings should take higher priority than previous setting
Then, the following conditions should be met:
- with 1, worker-num = 4
- with 1,2 worker-num = 2
- with 1,2,3, or 1, 3  worker-num = 1

siyuan0322 (approved):
How to specify the value by `g.with(pegasus.worker.num).V()`?
Or it should be `g.with("pegasus.worker.num", 2).V()`?

interactive_engine/compiler/src/main/java/com/alibaba/graphscope/gremlin/plugin/processor/IrStandardOpProcessor.java
shirly121: make pegasus.XX settings configurable per query:
```
g.with(pegasus.worker.num).V()
g.with(pegasus.batch.size).V()
g.with(pegasus.output.capacity).V()
```
the priority is : query > system env > system property > config file

Issue #3327, opened 2023-11-06
[BUG] Connecting via Gremlin-Console with doc has some ERROR


JackyYangPassion:
**Describe the bug**
```
gremlin> :remote connect tinkerpop.server conf/remote.yaml

Error during 'connect' - Could not establish serializer - org.apache.tinkerpop.gremlin.driver.ser.GryoMessageSerializerV1d0
```
**To Reproduce**

1. config remote.yaml 

`serializer: { className: org.apache.tinkerpop.gremlin.driver.ser.GryoMessageSerializerV1d0, config: { serializeResultToString: true }}`

2. gremlin> :remote connect tinkerpop.server conf/remote.yaml

3. See error
Error during 'connect' - Could not establish serializer - org.apache.tinkerpop.gremlin.driver.ser.GryoMessageSerializerV1d0

**Screenshots**
![image](https://github.com/alibaba/GraphScope/assets/13795366/2901fe35-3aff-403d-81c2-378ebc8c5e24)

**Environment (please complete the following information):**
 - GraphScope version: v0.24.0
 - OS: MacOS

**Additional context**

PR #3328, opened 2023-11-06
Update tinkerpop_gremlin.md

JackyYangPassion:
fix error

## What do these changes do?
fix GIE doc

## Related issue number

#3327 

Fixes


longbinlai:
@JackyYangPassion Hi, it seems like there is no need to modify the default configuration in GraphScope. Simply modify the configuration at the client (your own gremlin client)

PR #3357, opened 2023-11-16
fix(interactive): extension use hash(srcId, dstId, edgeLabelId, edgePks) instead of increase eid

bufapiqi:
## What do these changes do?

extension use hash(srcId, dstId, edgeLabelId, edgePks) to generate eid.
It can support edge update scenarios where eid is not known in advance.
Compatible with the original way of generating eid, use "enable.hash.generate.eid parameter" to judge
if enable.hash.generate.eid parameter=true:
    use hash generate
else:
    use increase generate

## Related issue number

Fixes


siyuan0322 (approved):
Well down! 👍

PR #3359, opened 2023-11-16
fix(interactive): fix the bug in filtering after intersection in GIE Runtime in distributed computation

### LINKED ISSUE #3360: [BUG] unexpected results of filtering after intersection in GIE Runtime in distributed computation
**Describe the bug**

When starting GIE with 2 servers, the following query outputs unexpected results (on LDBC graph):

```
Match (comment:COMMENT|POST)-[:HASCREATOR]->(person:PERSON),
              (comment:COMMENT|POST)-[:HASTAG]->(tag:TAG),
              (person:PERSON) -[:HASINTEREST]->(tag:TAG)
        Return count(person);
```

The bug occurs when we do some filtering on the intersected vertices in the pattern.

BingqingLyu:
## What do these changes do?

This pr:
1) fix the bug in filtering after intersection in distributed computation;
2) optimize the case of filtering by label only that doesn't necessary to query the storage, in Auxilia operator.

## Related issue number

Fixes #3360

### DIFF
--- a/flex/engines/hqps_db/database/adj_list.h
+++ b/flex/engines/hqps_db/database/adj_list.h
@@ -114,12 +114,10 @@ class SinglePropGetter {
   using value_type = T;
   static constexpr size_t prop_num = 1;
   SinglePropGetter() {}
-  SinglePropGetter(std::shared_ptr<TypedRefColumn<T>> c) : column(c) {
-    CHECK(column.get() != nullptr);
-  }
+  SinglePropGetter(std::shared_ptr<TypedRefColumn<T>> c) : column(c) {}
 
   inline value_type get_view(vid_t vid) const {
-    if (vid == NONE) {
+    if (vid == NONE || column == nullptr) {
       return NullRecordCreator<value_type>::GetNull();
     }
     return column->get_view(vid);
@@ -149,15 +147,25 @@ class MultiPropGetter {
     if (vid == NONE) {
       return NullRecordCreator<result_tuple_t>::GetNull();
     }
-    return get_view(vid, std::make_index_sequence<sizeof...(T)>());
+    result_tuple_t ret;
+    fill_result_tuple(ret, vid);
+    return ret;
   }
 
-  template <size_t... Is>
-  inline result_tuple_t get_view(vid_t vid, std::index_sequence<Is...>) const {
-    if (vid == NONE) {
-      return NullRecordCreator<result_tuple_t>::GetNull();
+  template <size_t I = 0>
+  inline typename std::enable_if<I == sizeof...(T), void>::type
+  fill_result_tuple(result_tuple_t& ret, vid_t vid) const {}
+
+  template <size_t I = 0>
+  inline typename std::enable_if<(I < sizeof...(T)), void>::type
+  fill_result_tuple(result_tuple_t& ret, vid_t vid) const {
+    using cur_ele_t = typename std::tuple_element<I, result_tuple_t>::type;
+    if (std::get<I>(column) == nullptr) {
+      std::get<I>(ret) = NullRecordCreator<cur_ele_t>::GetNull();
+    } else {
+      std::get<I>(ret) = std::get<I>(column)->get_view(vid);
     }
-    return std::make_tuple(std::get<Is>(column)->get_view(vid)...);
+    fill_result_tuple<I + 1>(ret, vid);
   }
 
   inline MultiPropGetter<T...>& operator=(const MultiPropGetter<T...>& d) {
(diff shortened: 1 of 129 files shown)


interactive_engine/executor/ir/graph_proxy/src/apis/graph/mod.rs
```diff
@@ -174,4 +174,12 @@ impl QueryParams {
             None
         }
     }
+
+    // is_queryable doesn't consider tables as we assume that the table info can be inferred directly from current data.
+    pub fn is_queryable(&self) -> bool {
+        !(self.filter.is_none()
+            && self.limit.is_none()
```
longbinlai, line 181: Why we need to consider limit and sample_ratio then. And if this is the case, directly using `! ( self.filter.is_none() && self.columns.is_none())` without introducing this arbitrary function is a better idea.

BingqingLyu: Done

flex/engines/hqps_db/database/adj_list.h
```diff
@@ -114,12 +114,10 @@ class SinglePropGetter {
   using value_type = T;
   static constexpr size_t prop_num = 1;
   SinglePropGetter() {}
-  SinglePropGetter(std::shared_ptr<TypedRefColumn<T>> c) : column(c) {
-    CHECK(column.get() != nullptr);
-  }
+  SinglePropGetter(std::shared_ptr<TypedRefColumn<T>> c) : column(c) {}
 
   inline value_type get_view(vid_t vid) const {
-    if (vid == NONE) {
```
longbinlai, line 122: CAN use columns.is_nullptr()

interactive_engine/executor/ir/runtime/src/process/operator/map/get_v.rs
```diff
@@ -159,17 +159,22 @@ impl FilterMapFunction<Record, Record> for AuxiliaOperator {
             // e.g., for g.V().out().as("a").has("name", "marko"), we should compile as:
             // g.V().out().auxilia(as("a"))... where we give alias in auxilia,
             //     then we set tag=None and alias="a" in auxilia
-            // 1. filter by labels.
+
+            // 1. If to filter by labels, and the entry itself carries label information already, directly eval it without query the store
```
longbinlai, line 163: Can the following code be replaced with params.ha_labels()?

interactive_engine/executor/ir/runtime/src/process/operator/map/get_v.rs
```diff
@@ -159,17 +159,22 @@ impl FilterMapFunction<Record, Record> for AuxiliaOperator {
...
             if !self.query_params.labels.is_empty() && entry.label().is_some() {
                 if !self
                     .query_params
                     .labels
                     .contains(&entry.label().unwrap())
                 {
+                    // pruning by labels
                     return Ok(None);
+                } else if self.query_params.filter.is_none() && self.query_params.columns.is_none() {
```
longbinlai, line 172: Use has_predicates & has_columns?

interactive_engine/executor/ir/core/src/plan/logical.rs
```diff
@@ -1477,7 +1477,13 @@ fn is_whole_graph(operator: &pb::logical_plan::Operator) -> bool {
                     && scan
                         .params
                         .as_ref()
-                        .map(|params| !params.is_queryable() && is_params_all_labels(params))
+                        .map(|params| {
+                            !(params.has_columns()
+                                || params.has_predicates()
+                                || params.has_sample()
+                                || params.has_limit())
+                                && is_params_all_labels(params)
```
longbinlai, line 1485: Can is_params_all_labels be replaced with params.has_labels()?

interactive_engine/executor/ir/runtime/src/process/operator/map/get_v.rs
BingqingLyu: Done.

interactive_engine/executor/ir/runtime/src/process/operator/map/get_v.rs
BingqingLyu: Done

interactive_engine/executor/ir/core/src/plan/logical.rs
BingqingLyu: Maybe not. Here we have to confirm it is the "whole graph", that it either specifies all labels, or no labels (also indicates that all labels is allowed).

longbinlai (approved):
LGTM

PR #3372, opened 2023-11-22
fix(interactive): Add a new release-with-debug-info profile

siyuan0322:
Also upgrades tonic version to fix warnings.


longbinlai (approved):
LGTM

Issue #3177, opened 2023-09-04
[BUG] Some bugs in `count()` in Pegasus Engine


BingqingLyu:
**Describe the bug**
As titled.

**To Reproduce**
**Query 1:**
```
#[test]
fn limit_filtermap_flatmap_count_test() {
    let mut conf = JobConf::new("limit_filtermap_flatmap_count_test");
    conf.set_workers(2);
    let mut result = pegasus::run(conf, || {
        |input, output| {
            input
                .input_from(1..50000000u32)?
                .limit(1000000)?
                .filter_map(|i| Ok(Some(i)))?
                .flat_map(|x| Ok(vec![x + 1].into_iter()))?
                .count()?
                .sink_into(output)
        }
    })
    .expect("build job failure");

    while let Some(v) = result.next() {
        println!("get {}", v.unwrap());
    }
}
```

it outputs error message as follows:
```
[reactor 2] [pegasus/src/data_plane/intra_process.rs:41] [worker_0(12526101200027122362)]: IntraProcessPush#push: send data failure Event { from_worker: 0, target_port: (2.0), kind: Cancel([root]) }
[reactor 2] [pegasus/src/communication/input/input.rs:398] [worker_0(12526101200027122362)]: propagate cancel of [root] failure, error: IOError(kind=System(NotConnected)),    occurred at: pegasus/src/data_plane/intra_process.rs:43, ;
```

**Query 2:**

```
#[test]
fn iterate_emit_flatmap_test() {
    let mut conf = JobConf::new("iterate_emit_flatmap_test");
    conf.set_workers(2);
    let mut result = pegasus::run(conf, || {
        |input, output| {
            input
                .input_from(1..50000000u32)?
                .limit(1000000)?
                .filter_map(|i| Ok(Some(i)))?
                .iterate_emit_until(IterCondition::max_iters(1), EmitKind::Before, |start| {
                    Ok(start
                        .flat_map(|x| Ok(vec![x + 1].into_iter()))?
                        .repartition(|x| Ok((x % 32) as u64)))
                })?
                .flat_map(|x| Ok(vec![x + 1].into_iter()))?
                .count()?
                .sink_into(output)
        }
    })
    .expect("build job failure");

    while let Some(v) = result.next() {
        println!("get {}", v.unwrap());
    }
}
```
It outputs 2  count values, i.e., #worker_num of count value

PR #3380, opened 2023-11-27
fix(interactive): event message error

lnfjpt:
## Related issue number

Fixes
https://github.com/alibaba/GraphScope/issues/3177


longbinlai (commented):
I would recommend a new status (instead of reusing the NotReady status) for worker, indicating that this worker has finished its own job but just been waiting for the others to complete. How about, "AwaitingFinSync" ?

PR #3418, opened 2023-12-12
feat(interactive): Support implicit type conversion for vertex index scan

zhanglei1949:
For Interactive compute engine, we need to provide the ability for implicit type conversion in predicate. (The only scenario where the type conversion may happen).

For most of the predicates, we convert it to a c++ expression, and the implicit conversion if down by C++ compiler. For example.
```cypher
MATCH(p : Person) where p.creditCardId = 1234 return p;
```
will be convert into an expression
```c++
bool predicate(int64_t creditCardId, int32_t target_value) {
    return creditCardId == target_value; // compare between a int64_t and int32_t.
}
```

However, when the query perform index scan for the vertex, we use indices for faster scan.

```cypher
MATCH(p: Person { id : 1234}) return p;
```

will be compiled to 
```c++
auto vertexSet = graph.ScanVertexFromOid(person_label_id, Any::From(1234));
```

Which will cause runtime error.

So, In this PR, we enable `id_indexer` support implicit type conversion between currently supported primary keys:
- uint32_t
- uint64_t
- int32_t
- uint32_t

Note that conversion between String and other type are not supported, since compiler will throw the error before compiling to a physical plan.

### DIFF
--- a/flex/utils/id_indexer.h
+++ b/flex/utils/id_indexer.h
@@ -142,6 +142,77 @@ struct KeyBuffer<std::string_view> {
   }
 };
 
+// Implicit conversion for OID_T
+// For example, we need to convert int64_t to int32_t, or int32_t to int64_t.
+struct ConvertHelper {
+  static inline bool convert(const Any& from, PropertyType to_type, Any& to) {
+    // currently only support convert primary key types.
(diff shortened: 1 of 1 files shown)


zhanglei1949:
Not going to be merged since we expect compiler to do the type conversion/check.

Issue #3399, opened 2023-12-05
refactor(interactive): Refactor the `AbstractArrowFragmentLoader` for better concurrency


zhanglei1949:
Currently, graph loading is done concurrently at the granularity of files. However, there is a significant size difference between files. We should perform concurrency at the granularity of `RecordBatch`.

zhanglei1949:
Also, `ODPSFragment` can refactored to allow parallel stream reading, since it allows splitting table by rows.

zhanglei1949:
Closed since already implemented.

PR #3468, opened 2024-01-03
refactor(interactive): Refactor `AbstractFragmentLoader` for RecordBatch parallelism

zhanglei1949:
Fix issue #3399 

The BulkLoading step can be divided into two part

- Read vertex/edge table from file/ODPS to memory
- Construct graph structures(vertex_map, csr) in memory and save to files

In current implementation, it is hard for us to paralleling graph structure construction. BUT we can accelerate the raw data reading.

TODO: Parallelize graph structure construction.


zhanglei1949:
Closed since the parallel loading will be implemented via other PR.

PR #3532, opened 2024-02-04
fix(k8s): Bump up vineyard to v0.20.3

dashanji:
## What do these changes do?

The PR is to fix the bug caused by the version mismatch of vineyard in the graphscope-dev image.

As:
https://github.com/alibaba/GraphScope/actions/runs/7771225038/job/21192444777
https://github.com/alibaba/GraphScope/actions/runs/7753022954/job/21144025198

Also, the dev image updated.
https://github.com/alibaba/GraphScope/actions/runs/7773830300

### DIFF
--- a/.github/workflows/build-graphscope-wheels-macos.yml
+++ b/.github/workflows/build-graphscope-wheels-macos.yml
@@ -185,7 +185,7 @@ jobs:
       run: |
         . ~/.graphscope_env
         python3 -m pip install libclang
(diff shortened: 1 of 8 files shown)


siyuan0322 (commented):
Don't do this... this vineyard version is fixed within each release

dashanji:
@siyuan0322 Thanks for the advice, I update the dev image manually. Also, the `k8s-test` CI passed, could you please take another look at this?

siyuan0322 (approved):

PR #3630, opened 2024-03-13
fix(interactive): Fix Timeout Bug & Process Results From Pegasus Synchronously

shirly121:
## What do these changes do?
as titled.

## Related issue number

Fixes

PR #3647, opened 2024-03-20
fix(interactive): Fix CI Bug Caused By PR #3630

shirly121:
## What do these changes do?
The issue is caused by PR #3630, which changed the result collection from asynchronous to synchronous. After submitting a request to the client, it is necessary to actively call a request to retrieve the result from the buffer. However, the test chain forgot to add this call, leading to the problem. This PR mainly fix the issue.

## Related issue number

Fixes


zhanglei1949 (approved):
LGTM

siyuan0322 (approved):

PR #3650, opened 2024-03-23
feat(interactive): Enable starting compiler as a subprocess of interactive server

zhanglei1949:
As our `AdminService` is able to switching services between graphs, we should be also able to switch graph on compiler service. For now, we make `GraphServer` a subprocess of `InteractiveServer`.


luoxiaojian (approved):

LiSu (approved):
LGTM

PR #3677, opened 2024-03-28
fix(learning): Fix torch version  to be compatible with gcc8 and python3.7

LiSu:
Fix torch version at 1.13.1 to be compatible with gcc8 and python3.7


lidongze0629 (approved):
LGTM

husimplicity (approved):

PR #3723, opened 2024-04-18
fix(interactive): Identify pk when it is fused with other predicates in IRCore

### LINKED ISSUE #3722: [BUG] IRCore cannot recognize pk when it is fused with other predicates
**Describe the bug**

Currently,  IRCore cannot recognize pk when it is fused with other predicates. For example, for the Gremlin query:
```
g.V().hasLabel("PERSON").has("id",1).has("name","marko")
```
where "id" is the primary key of "PERSON".
Compiler will fuse it into a expression like `id==1 && name==marko`, and ir-core failed to identify that "id" as the pk.

BingqingLyu:
## What do these changes do?

As titled.

## Related issue number

Fixes #3722


longbinlai (approved):
This is a temporary solution, which will soon be replaced by the new Gopt framework. 
@BingqingLyu Can you add an issue for the Gopt replacement thing?

shirly121 (approved):

BingqingLyu:
> This is a temporary solution, which will soon be replaced by the new Gopt framework. @BingqingLyu Can you add an issue for the Gopt replacement thing?

ok

siyuan0322 (approved):

PR #3725, opened 2024-04-18
feat(interactive): Adapt to new unified schema

zhanglei1949:
Migrating the graph schema of Interactive to the Flex `Unified` graph schema.

Those changes are included in this PR.

- Graph, Vertex and Edge can have descriptions.
- Introduce `version` in schema, which denotes the version of the schema definition. If not given, `v0.0` is used. The schema of version `v0.0` is exactly the previous schema definition. Therefore, the schema of the graph used previously can continue to be used without any modifications.
- `default_value` and `nullable` are not supported in interactive currently.
- `type_id` and `property_id` shall not be provided by user, mutable_csr will generate the internal id instead.


zhanglei1949:
Currently we can not return vertex/edge types in stored procedure, so codegen of simple_match_10.cypher fails. We will support it in the future @shirly121 .

shirly121 (approved):

lidongze0629 (approved):

PR #3726, opened 2024-04-18
refactor(interactive): Rename two temporal types, Date and Day to TimeStamp and Date

zhanglei1949:
Currently Interactive support two temporal data types,
- Date
- Day

However, the naming of these two types doesn't really match its implementation. The struct `Date` actually stores a timestamp(in million second), the struct `Day` actually stores a Date(Year, Month and Day).

- This PR **will also change the** the data type usage in graph schema. 
- This PR may affect the previously written stored procedures, where `set_day` and `set_date` maybe used. Should be changed to `set_date` and `set_timestamp`.


zhanglei1949:
Closed since will not be merged

PR #3764, opened 2024-05-06
fix(interactive): Fix signal handling for interactive

### LINKED ISSUE #3760: [BUG] The compiler process seems not gracefully shutdown even if the interactive_server is down
When the Interactive server got killed, the subprocess which running the compiler should also be cleaned.

zhanglei1949:
- Fix the signal handling of `actor_system`, add signal handlers for `SIGINT` and `SIGTERM` to clean up actor_system and resources(including subprocess).
- Fix the route setting in `hqps_http_handler` and `admin_http_handler`.  Each match_rule should have its own handler pointer, can not be shared.
- Remove the handling of unused url `/interactive/query`
- Fix some destructors.

Fix #3760


zhanglei1949:
Thanks for @LiSu for help debugging. :)

liulx20 (dismissed):

LiSu (approved):

LiSu (approved):

liulx20 (approved):

PR #3800, opened 2024-05-15
fix(interactive): persist latest kafka queue offset before stopped

siyuan0322:


BingqingLyu (approved):
The configuration of `timeout` can be fixed in anther pr @lnfjpt .

PR #3816, opened 2024-05-20
chore: Add estimate memory usage command for gsctl

siyuan0322:
Example usage:
```bash
❯ PYTHONPATH=`pwd` python3 graphscope/gsctl/gsctl.py estimate -v 222111111 -e 1610000000 -vf 2 -ef 28 -g gae -p 4

Disclaimer: The `estimate` command serves as a estimator for various kind of GraphScope resources.
    - GAE memory usage estimator:
        It would do a rough estimation of the total memory usage of the GAE.
        The actual memory usage may vary vastly due to the complexity of the graph algorithm and the data distribution.

        Here's some assumption when estimating the memory usage:
            1. Assuming the graph is a simple graph, e.g. only has 1 label and at most 1 property for each label.
            2. Assuming the graph algorithms has a fixed amount of memory consumption, e.g. PageRank, SSSP, CC;
               It should not generate a huge amount of intermediate result, like K-Hop, Louvain.
            3. Assuming the vertex map is global vertex map

        User should take this estimation as a reference and a start point for tuning, and adjust the memory allocation according to the actual situation.

[SUCCESS] The estimated memory usage is 30.61 GB per pod for 4 pods.
```

### DIFF
--- a/python/graphscope/gsctl/commands/common.py
+++ b/python/graphscope/gsctl/commands/common.py
@@ -52,7 +52,7 @@ def connect(coordinator_endpoint):
         context = get_current_context()
         if context is None:
             err(
-                "No available context found, try to connect by `gsctl conenct --coordinator-endpoint <addr>`."
+                "No available context found, try to connect by `gsctl connect --coordinator-endpoint <addr>`."
             )
             return
         coordinator_endpoint = context.coordinator_endpoint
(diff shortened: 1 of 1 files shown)


python/graphscope/gsctl/commands/common.py
```diff
@@ -82,5 +82,70 @@ def close():
...
+    offset_array = v_num * 8 / gb
+    # edge table
+    edge_table = e_file_size
+    return vertex_map + vertex_table + edge + offset_array + edge_table
+
+
+@cli.command()
+@click.option(
+    "-s",
+    "--storage",
```
yecol, line 128: -e and --engine?

python/graphscope/gsctl/commands/common.py
```diff
@@ -82,5 +82,70 @@ def close():
...
+    type=click.Choice(["gae"], case_sensitive=False),
+    help="Storage type",
+    required=True,
+)
+@click.option("-v", "--v-num", type=int, help="Number of vertices")
+@click.option("-e", "--e-num", type=int, help="Number of edges")
+@click.option("-vf", "--v-file-size", type=float, help="Size of vertex files in GB")
+@click.option("-ef", "--e-file-size", type=float, help="Size of edge files in GB")
+def estimate(storage, v_num, e_num, v_file_size, e_file_size):
```
yecol, line 137: add an option:
-d, --distributed, e.g., by default 4 pods
return each pod configurations.

python/graphscope/gsctl/commands/common.py
```diff
@@ -82,5 +82,70 @@ def close():
...
+    2. edge
+        - CSR + CSC: 2 * #E * (sizeof(inner_id) + sizeof(edge_id))
+        - offset array: #V * sizeof(size_t)
+        - edge table: size of vertex files in uncompressed CSV format, unit is GB
+    3. additional data structure when graph is partitioned:
+        - outer vertex ID to local ID map:
+        - local ID to outer vertex ID array:
+
+This is the minimum memory usage of the graphscope engine with 1 partition, the actual memory usage would be larger than this estimation.
+User should enlarge it by a multiple it by some factor (e.g. 2) to simulate the memory usage of the graphscope engine.
```
yecol, line 106: Return the value by this factor. 
The value is assumed to be used directly.

python/graphscope/gsctl/commands/common.py
```diff
@@ -82,5 +82,70 @@ def close():
...
 
+disclaimer = """
+Disclaimer: The `estimate` command is a roughly estimation of the memory usage of the graphscope engine.
+The actual memory usage may vary vastly due to the complexity of the graph algorithm and the data distribution.
+
+The estimation is based on the following formulas:
+let #V = number of vertices, #E = number of edges,
+assume load_factor of hashmap is 0.41, sizeof(inner_id) = sizeof(edge_id) = sizeof(vertex_id) = 8 bytes,
+where vertex_id is the original representation of vertex, it could be string though, in this case, the memory usage will grow.
+Graph:
```
yecol, line 93: - Move this as comments of the function for internal reference. No need to explain in such detail. 
- Spell out the assumption when you estimate, not how to estimate....
     - Assuming the graph is a simple graph, without labels, properties, and features...
     - Assuming the graph algorithms you are going to run will not generate a huge amount of intermediate data, e.g., PageRank, SSSP, CC, not algorithms like Louvain...

python/graphscope/gsctl/commands/common.py
```diff
@@ -82,5 +82,70 @@ def close():
             info("Try 'gsctl --help' for help.")
 
 
+disclaimer = """
+Disclaimer: The `estimate` command is a roughly estimation of the memory usage of the graphscope engine.
```
yecol: `estimator`, and please disclaim it as resources estimator, rather than just for memory usage. 
It should give guide for (in the future):
- memory for gae
- disk usage for groot/interactive
- k8s config for graphscope one...

yecol (changes requested):

python/graphscope/gsctl/commands/common.py
siyuan0322: e is used for enum. Seems the usage is more relate to storage.

yecol (dismissed):

PR #3859, opened 2024-05-30
feat(interactive): Implement the `get_graph_statistics` for AdminService

zhanglei1949:
`curl -X GET http://localhost:7777/v1/graph/{graph_id}/statistics` retrieves the graph's stats: total vertex and edge counts, plus label counts for each vertex and edge.

Currently, this API will only returns the statistics of the graph if it is running. inquire about a non-running graph will yield an error.

### DIFF
--- a/flex/storages/metadata/graph_meta_store.h
+++ b/flex/storages/metadata/graph_meta_store.h
@@ -223,6 +223,28 @@ struct UpdateJobMetaRequest {
   static UpdateJobMetaRequest NewFinished(int rc);
 };
 
+struct GraphStatistics {
+  // type_id, type_name, count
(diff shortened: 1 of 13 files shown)


flex/storages/metadata/graph_meta_store.h
```diff
@@ -223,6 +223,28 @@ struct UpdateJobMetaRequest {
   static UpdateJobMetaRequest NewFinished(int rc);
 };
 
+struct GraphStatistic {
```
longbinlai, line 226: Should be `GraphStatistics`.

zhanglei1949: Fixed.

liulx20 (approved):

shirly121 (approved):

PR #3876, opened 2024-06-04
feat(interactive): Add field id and name in graph schema

zhanglei1949:
In flex schema definition, we have graph id and graph name. But in `schema` implementation we have no such field.
More other, having field `id` in `schema.h` will make it easy for us to get the current running graph's id.


flex/storages/rt_mutable_graph/schema.cc
```diff
@@ -1097,6 +1107,20 @@ static Status parse_schema_from_yaml_node(const YAML::Node& graph_node,
...
+    schema.set_name(graph_node["name"].as<std::string>());
+  } else {
+    LOG(ERROR) << "name is not set properly";
+    return Status(StatusCode::InvalidSchema, "grpah name is not set properly");
+  }
+
+  if (graph_node["id"]) {
+    schema.set_id(graph_node["id"].as<std::string>());
+  } else {
+    LOG(WARNING) << "id is not set properly, use name as id";
```
zhanglei1949, line 1120: If id is not specified, we use name as id.

zhanglei1949:
Closed since not going to be adapted.

PR #3887, opened 2024-06-06
fix(interactive): Support add property with exist label

bufapiqi:
## What do these changes do?
support add label properties with exist label when system is running

## Related issue number

Fixes


interactive_engine/executor/store/groot/src/db/graph/types/vertex.rs
```diff
@@ -131,6 +131,33 @@ impl VertexTypeManager {
         Ok(())
     }
 
+    pub fn update_type(
+        &self, si: SnapshotId, label: LabelId, codec: Codec, table0: Table,
+    ) -> GraphResult<()> {
+        assert_eq!(si, table0.start_si, "type start si must be equal to table0.start_si");
```
siyuan0322, line 137: Should return Err. assert_eq would cause panic

siyuan0322 (dismissed):

PR #3890, opened 2024-06-06
fix(interactive): Unified Query Cache for Gremlin and Cypher Queries

shirly121:
## What do these changes do?

1. Use a unified query cache for both Gremlin and Cypher queries.
2. Use the pre-optimization `LogicalPlan` as the query cache key to avoid applying optimizations when computing the cache key.
3. Directly use the string from the plan explanation as the hash code, instead of computing it from all attributes within each operator of the plan. This is mainly due to the following reasons:
    1. Avoid modifying the hash code each time when a new attribute is added to an operator.
    2. For some static attributes, Java does not allow overloading, and it does not guarantee these attributes will remain unchanged after a service restart, such as [enum](https://stackoverflow.com/questions/19496751/enum-hash-code-calculation-are-not-the-same).

## Related issue number

Fixes


zhanglei1949 (approved):
LGTM

PR #3912, opened 2024-06-12
docs(interactive): Add document about `column_mapping`

zhanglei1949:
Clarifying the behavior of interactive when importing data from diverse sources.


docs/flex/interactive/data_import.md
```diff
@@ -6,6 +6,28 @@ In our guide on [using custom graph data](./custom_graph_data.md), we introduced
...
+    index: 0  # Column index in the data source
+    name: col_name # If a column name is present
+  property: property_name  # The mapped property name
+```
+
+The column mapping requirements differ based on the data source:
+
+#### Import from CSV 
+
+You can provide either `index`, `name`, or both.
```
lidongze0629, line 25: which one has higher priority if `index` and `name` exist simultaneously ?

zhanglei1949: Updated in documentation.
If Both name and index are specified, we will check whether they points to the same column.

lidongze0629 (approved):
LGTM.

docs/flex/interactive/data_import.md
longbinlai: What if they are not?

zhanglei1949: If not, exception will be returned.

liulx20 (approved):

PR #3929, opened 2024-06-17
fix(interactive): Fixed bug that wasted disk space

yqylh:
## What do these changes do?
The interactive service generates n cache files named neo4j-xxx in the /tmp/ directory during execution, each with a size of at least **0.5GB**. These files do not get deleted when the process ends. Here, n represents the number of times the interactive server has been started or restarted.

This issue was due to the previously used termination method for the compiler process, which utilized the **terminate()** function. Since terminate() does not facilitate a normal exit, it led to the compiler's inability to clean up the cache files it created.

Transitioning to sending the **SIGINT signal** ensures that the compiler exits gracefully and performs the necessary cache cleanup. And sleep for up to 10 seconds to wait for the compiler process to stop


zhanglei1949 (approved):
LGTM

liulx20 (approved):

PR #3934, opened 2024-06-17
fix(interactive): Avoid redundant traversal when building lf_indexer

liulx20:
- Avoid redundant traversal when building the lf_indexer

- Return an invalid value to indicate the key doesn't exist, instead of crashing.


zhanglei1949 (approved):
LGTM

luoxiaojian (approved):

zhanglei1949 (approved):
LGTM

Issue #3947, opened 2024-06-18
Support processing `None` value in GIE


BingqingLyu:
**Is your feature request related to a problem? Please describe.**

As titled.

BingqingLyu:
Supported in #3972

PR #3972, opened 2024-06-25
fix(interactive): Support NULL Value in expression in GIE Runtime

BingqingLyu:
## What do these changes do?

As titled.

## Related issue number

#3947


longbinlai (approved):
The docs of handling NULL will be given after Interactive has implemented the same semantics.

zhanglei1949 (approved):

PR #3990, opened 2024-07-01
refactor(interactive): Support customizing ports when deploy interactive in local mode and connect via SDK

zhanglei1949:
Interactive service exposes many ports, including `admin_port`, `query_port(stored_proc_port)`, `cypher_port` and `gremlin_port`. In previous SDK implementation, we create a driver by only providing the `admin_endpoint`, hoping for get all other ports via `GET /v1/service/status`, However, this only works if the SDK runs inside the service's container. In cases like k8s deployment, the internal port will be mapped to a large port number which is chosen by the cluster.

In this PR, we let user to declare the endpoints to environment variables, and connect via SDK.
```bash
export INTERACTIVE_ADMIN_ENDPOINT: http://host:port
export INTERACTIVE_STORED_PROC_ENDPOINT: http://host:port
export INTERACTIVE_CYPHER_ENDPOINT: neo4j://host:port or bolt://host:port
export INTERACTIVE_GREMLIN_ENDPOINT: ws://host:port/gremlin
```

For example, with python sdk, user can connect to the Interactive service after all environment variable is exported.
```python
from gs_interactive.client.driver import Driver
driver = Driver()
with driver.session() as sess:
   #....
```

The documentaion will be added after the interactive documentation is refactored with `gsctl`

### DIFF
--- a/flex/interactive/sdk/java/src/main/java/com/alibaba/graphscope/interactive/client/Driver.java
+++ b/flex/interactive/sdk/java/src/main/java/com/alibaba/graphscope/interactive/client/Driver.java
@@ -32,36 +32,96 @@ import java.util.logging.Logger;
 public class Driver {
     private static final Logger logger = Logger.getLogger(Driver.class.getName());
 
-    private final String host;
-    private final int port;
+    private final String adminUri;
+    private final String storedProcUri;
+    private String cypherUri;
+    private String gremlinUri;
+    private String host;
+    private int port;
     private Session defaultSession;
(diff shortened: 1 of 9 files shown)


liulx20 (dismissed):

longbinlai (approved):
LGTM

lidongze0629 (approved):

lidongze0629 (approved):

liulx20 (approved):

PR #4024, opened 2024-07-09
fix(interactive): Add graph_id/proc_id type check for python sdk

zhanglei1949:
As titled.

### DIFF
--- a/flex/interactive/sdk/python/gs_interactive/client/session.py
+++ b/flex/interactive/sdk/python/gs_interactive/client/session.py
@@ -438,7 +438,7 @@ class DefaultSession(Session):
         self,
         graph_id: Annotated[StrictStr, Field(description="The id of graph to get")],
     ) -> Result[GetGraphSchemaResponse]:
-        self.ensure_param_str("graph_id", graph_id)
+        graph_id = self.ensure_param_str("graph_id", graph_id)
         try:
             response = self._graph_api.get_schema_with_http_info(graph_id)
             return Result.from_response(response)
@@ -449,7 +449,7 @@ class DefaultSession(Session):
         self,
         graph_id: Annotated[StrictStr, Field(description="The id of graph to get")],
     ) -> Result[GetGraphResponse]:
-        self.ensure_param_str("graph_id", graph_id)
+        graph_id = self.ensure_param_str("graph_id", graph_id)
         try:
             response = self._graph_api.get_graph_with_http_info(graph_id)
             return Result.from_response(response)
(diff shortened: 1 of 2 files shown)


flex/interactive/sdk/python/gs_interactive/client/session.py
```diff
@@ -814,4 +829,11 @@ def try_upload_files(self, schema_mapping: SchemaMapping) -> Result[SchemaMappin
             return upload_res
         print("new schema_mapping: ", upload_res.get_value())
         return Result.ok(upload_res.get_value())
+
+    def ensure_param_str(self, param_name : str, param):
```
longbinlai, line 833: 这个如果是int之类的类型，可以调用str(x) 尝试转string。避免这种比较简单的情况也抛异常

flex/interactive/sdk/python/gs_interactive/client/session.py
```diff
@@ -814,4 +830,13 @@ def try_upload_files(self, schema_mapping: SchemaMapping) -> Result[SchemaMappin
             return upload_res
         print("new schema_mapping: ", upload_res.get_value())
         return Result.ok(upload_res.get_value())
-        
\ No newline at end of file
+
+    def ensure_param_str(self, param_name : str, param):
```
longbinlai, line 834: 这个这么写好了：
```
ret = param
if not isinstance(param, str):
    try:
        ret = str(param)
    except e:
        raise  ValueException(f"{param_name} has a non-string value {param}, and cannot be converted into a string")

return ret
```

flex/interactive/sdk/python/gs_interactive/client/session.py
```diff
@@ -560,6 +572,7 @@ def get_procedure(
     def call_procedure(
         self, graph_id: StrictStr, params: QueryRequest
     ) -> Result[CollectiveResults]:
+        graph_id = self.ensure_param_str("graph_id", graph_id)
```
longbinlai, line 575: 这里要把 `self.ensure_param_str` 放到try里面，这个exception才能被捕捉到

zhanglei1949: The `try catch` block catches errors from the server side and encapsulates them in a `Result`. For parameter type checking, it is preferable to throw them directly to the user.

flex/interactive/sdk/python/gs_interactive/client/session.py
zhanglei1949: I'm considering whether converting any object to a string is advisable in our situation. For instance, we convert an integer value to a string because it's a frequent mistake when using the Interactive SDK. Concerning other types, perhaps it's more appropriate to trigger the error on the client side instead of sending a request to the server knowing that it will fail.

lidongze0629 (approved):

longbinlai (approved):

PR #4031, opened 2024-07-11
fix(interactive): Support parsing quoting_char when reading header row from csv

zhanglei1949:
As titled.


flex/storages/rt_mutable_graph/loader/csv_fragment_loader.cc
```diff
@@ -110,6 +112,13 @@ static std::vector<std::string> read_header(const std::string& file_name,
       while (std::getline(ss, token, delimiter)) {
         // trim the token
         token.erase(token.find_last_not_of(" \n\r\t") + 1);
+        // trim the quote char at the beginning and end of the token
```
longbinlai, line 115: I have asked GPT  to further refine these line of code and got:
```
// Function to trim whitespace from both ends of a string
std::string trim(const std::string& str) {
    const std::string whitespace = " \n\r\t";
    const auto start = str.find_first_not_of(whitespace);
    const auto end = str.find_last_not_of(whitespace);
    
    return (start == std::string::npos || end == std::string::npos)
           ? "" : str.substr(start, end - start + 1);
}

// Function to remove surrounding quotes from a string if present
std::string remove_surrounding_quotes(const std::string& str, char quote_char) {
    if (str.size() >= 2 && str.front() == quote_char && str.back() == quote_char) {
        return str.substr(1, str.size() - 2);
    }
    return str;
}

// Function to unescape escaped quotes inside a string
std::string unescape_quotes(const std::string& str, char quote_char) {
    std::string result;
    result.reserve(str.size());

    for (size_t i = 0; i < str.size(); ++i) {
        if (str[i] == '\\' && i + 1 < str.size() && str[i + 1] == quote_char) {
            result.push_back(quote_char);
            ++i; // Skip the escaped quote
        } else {
            result.push_back(str[i]);
        }
    }
    return result;
}

// Function to process the token
std::string process_token(const std::string& token, char quote_char) {
    std::string trimmed_token = trim(token);
    std::string unquoted_token = remove_surrounding_quotes(trimmed_token, quote_char);
    std::string final_token = unescape_quotes(unquoted_token, quote_char);
    return final_token;
}
```

zhanglei1949: Improved the code derived from GPT without fully using it.

liulx20 (approved):

longbinlai (approved):

PR #4067, opened 2024-07-18
fix(analytical): involving triangles

siyuan0322:


siyuan0322:
turns out we have some CI replies on the triangles.

PR #4131, opened 2024-08-13
fix(interactive): fix zero property edge and schema check

yqylh:
The check of the schema reports an error when adding, deleting, or checking with zero attribute edge.
rt_server can't get metadata_store_, so metadata_store_ is empty.


flex/engines/graph_db/database/graph_db_operations.cc
```diff
@@ -249,12 +253,15 @@ EdgeData GraphDBOperations::inputEdge(const nlohmann::json& edge_json,
   edge.src_pk_value = Any(jsonToString(edge_json["src_primary_key_value"]));
   edge.dst_pk_value = Any(jsonToString(edge_json["dst_primary_key_value"]));
   // Check that all parameters in the parameter
-  if (edge_json["properties"].size() != 1) {
+  if (edge_json["properties"].size() > 1) {
```
zhanglei1949, line 256: Could we support multiple properties on edge, in graph_db_operations?

yqylh: The storage and schema both support multi-attribute edges. However, all transactions involving edges and the underlying storage interfaces currently only support single-attribute edges. This should be a major project in the future.

luoxiaojian (approved):

zhanglei1949 (approved):
LGTM

PR #4171, opened 2024-08-21
fix(interactive):   vertex with string oid is not support

liulx20:
as titled.


flex/engines/graph_db/runtime/adhoc/operators/scan.cc
```diff
@@ -150,6 +150,81 @@ bool parse_idx_predicate(const algebra::IndexPredicate& predicate,
...
+      } else if (dt == common::DataType::STRING) {
+        std::string name = p.name();
+        std::string value = params.at(name);
+        oids.emplace_back(Any::From(value));
+      } else if (dt == common::DataType::INT32) {
+        std::string name = p.name();
+        std::string value = params.at(name);
+        int32_t v = std::stoi(value);
+        oids.emplace_back(v);
+      }
```
zhanglei1949, line 222: Need error handling here?

flex/engines/graph_db/runtime/adhoc/operators/scan.cc
```diff
@@ -150,6 +150,81 @@ bool parse_idx_predicate(const algebra::IndexPredicate& predicate,
   return true;
 }
 
+bool parse_idx_predicate(const algebra::IndexPredicate& predicate,
+                         const std::map<std::string, std::string>& params,
+                         std::vector<Any>& oids, bool& scan_oid) {
```
zhanglei1949, line 155: Cypher queries scan one oid at a time, while Gremlin allows specifying multiple oids in an array. We might implement support for a vector of oids, treating single id scans as a special case.

flex/engines/graph_db/runtime/adhoc/operators/scan.cc
liulx20: fix

longbinlai (approved):

siyuan0322 (approved):

PR #4199, opened 2024-09-02
fix(interactive): Fix Bugs of Type Inference in `both()`

shirly121:
## What do these changes do?
as titled.

## Related issue number

Fixes


BingqingLyu (approved):
LGTM.

PR #4220, opened 2024-09-11
feat(interactive): Introduce another example graph `graphscope-github-network` 

zhanglei1949:
Introduce a new example graph that illustrates the relationships between GitHub repositories, other repositories, and GitHub users.


zhanglei1949:
TODO @zhanglei1949 : We should put the csv files to gstest. not in this repo.

Issue #4243, opened 2024-09-19
[BUG] There seems to be a bug in library dependencies


zhanglei1949:
The dependency relationship is still unclear and confusion. When I add a test in `flex/tests/hqps/` with name `runtime_test.cc`, with following content
```c++
#include "flex/engines/graph_db/runtime/adhoc/runtime.h"
#include <map>
#include <string>
#include <vector>

int main(int argc, char** argv) {
  physical::PhysicalPlan plan;
  std::map<std::string, std::string> params;

  // Receiving output
  std::vector<char> buf;
  gs::Encoder output(buf);

  gs::runtime::Context ctx;
  ctx.append_tag_id(0);
  LOG(INFO) << "Finish query, output size: " << buf.size();

  return 0;
}
```

The project fail to build at linking time.

```txt
/usr/bin/ld: ../../engines/graph_db/runtime/libruntime_common.so: undefined reference to `gs::ReadTransaction::graph() const'
/usr/bin/ld: ../../engines/graph_db/runtime/libruntime_common.so: undefined reference to `gs::Schema::has_multi_props_edge() const'
/usr/bin/ld: ../../engines/graph_db/runtime/libruntime_common.so: undefined reference to `gs::MutablePropertyFragment::get_ie_csr(unsigned char, unsigned char, unsigned char) const'
/usr/bin/ld: ../../engines/graph_db/runtime/libruntime_common.so: undefined reference to `gs::Schema::get_vertex_property_names[abi:cxx11](unsigned char) const'
/usr/bin/ld: ../../engines/graph_db/runtime/libruntime_common.so: undefined reference to `gs::ReadTransaction::edge_iterator::GetNeighbor() const'
/usr/bin/ld: ../../engines/graph_db/runtime/libruntime_common.so: undefined reference to `gs::ReadTransaction::GetVertexIndex(unsigned char, gs::Any const&, unsigned int&) const'
/usr/bin/ld: ../../engines/graph_db/runtime/libruntime_common.so: undefined reference to `gs::ReadTransaction::GetInEdgeIterator(unsigned char, unsigned int, unsigned char, unsigned char) const'
/usr/bin/ld: ../../engines/graph_db/runtime/libruntime_common.so: undefined reference to `gs::ReadTransaction::edge_iterator::~edge_iterator()'
[...]

PR #4244, opened 2024-09-19
refactor(interactive): Make runtime module header only

zhanglei1949:
Move all implementation code from `.cc` to `.h` for the runtime_adhoc module.

- Reason for refactor: We aim to make the operator's implementation header-only and templatized, enabling other storage systems to easily run ad-hoc queries.
- Cost: Converting the runtime module to header-only may slow down compilation?

Fix Issue #4243


zhanglei1949:
Closed since will not be merged.

PR #4284, opened 2024-10-11
refactor(interactive): Resolve the dependency problem between compiler and interactive SDK

zhanglei1949:
The Interactive SDK and compiler both rely on gaia-ir-proto, including the protoc-generated classes. Previously, we used the maven-shade-plugin to shade the protoc files from the Interactive SDK, but this led to other build issues, as noted in PR #4255.

In this PR, we address the problem by adding an additional jar classifier, `no-gaia-ir`, which can be easily utilized by the compiler.


zhanglei1949:
TODO: We need to wait Interactive Java SDK 0.4.2 available in maven central repo, to pass the CI check.

siyuan0322 (approved):

shirly121 (approved):

PR #4301, opened 2024-10-25
feat(interactive): Introduce kafka as another wal storage

zhanglei1949:
Introduce `KafkaWalWriter` as another writer for persisting wals.

- [x] Add the documentation, including the performance comparison and usage, configuration.
- [x] Add CI tests.

The performance of `KafkaWalWriter` and `KafkaWalConsumer` could be improved, and will be addressed in the nea future,  see #4305


flex/engines/graph_db/database/wal.cc
```diff
@@ -70,28 +198,82 @@ void WalWriter::append(const char* data, size_t length) {
   file_used_ += length;
 
   if (static_cast<size_t>(write(fd_, data, length)) != length) {
-    LOG(FATAL) << "Failed to write wal file " << strerror(errno);
+    LOG(ERROR) << "Failed to write wal file " << strerror(errno);
```
zhanglei1949, line 201: The `append()` is refactored into returning a bool value, and check outside the function.

flex/engines/graph_db/database/graph_db_session.cc
```diff
@@ -251,6 +251,73 @@ AppBase* GraphDBSession::GetApp(int type) {
 
 #undef likely  // likely
 
+Result<std::string> GraphDBSession::IngestWals(const char* input, size_t len) {
```
zhanglei1949, line 254: Process the wal-ingest request

flex/engines/http_server/actor/executor.act.h
```diff
@@ -48,6 +48,7 @@ class ANNOTATION(actor:impl) executor : public hiactor::actor {
 
   seastar::future<admin_query_result> ANNOTATION(actor:method) get_edge(graph_management_query_param&& param);
 
+  seastar::future<admin_query_result> ANNOTATION(actor:method) ingest_wal(query_param&& param);
```
zhanglei1949, line 51: The `ingest_wal` API doesn't need to guarded by `BUILD_KAFKA_WAL_WRITER`

PR #4332, opened 2024-11-21
fix(interactive): Introduce `MetricsTool` to profile memory usage, pending tasks and qps

shirly121:
## What do these changes do?
the `MetricsTool` will print the following metrics periodically (setting of `metrics.tool.interval.ms`, default is 5 mins):
1. `memory.usage`: memory usage of jvm and direct memory
2. `rpc.channels.executor.queue`: pending tasks in each grpc (netty) channel
3. `gremlin.executor.queue`: pending tasks in gremlin executor
4. `gremlin.qps`: gremlin qps

These metrics will be logged in a separate file, configured by `PerfMetricLog` in [logback.xml](https://github.com/shirly121/GraphScope/blob/ir_metric_tool/interactive_engine/assembly/src/conf/groot/logback.xml#L37)

## Related issue number

Fixes

### DIFF
--- a/interactive_engine/assembly/src/conf/groot/logback.xml
+++ b/interactive_engine/assembly/src/conf/groot/logback.xml
@@ -34,6 +34,23 @@
         </encoder>
     </appender>
 
+    <appender name="PerfMetricLog" class="ch.qos.logback.core.rolling.RollingFileAppender">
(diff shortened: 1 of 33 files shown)


zhanglei1949 (approved):
LGTM

BingqingLyu (approved):

PR #4359, opened 2024-12-12
fix(interactive): Cut off the string if it exceeds max_length.

zhanglei1949:
Fix two bugs:
- Vertex table does not resize to actual size of the vertex record batch. #4383
- The string is not cut off when the length exceeds the specified max_length for `var_char`. #4365


zhanglei1949:
TODO: We should support VAR_CHAR on compiler first or all. @shirly121

zhanglei1949:
> compiler first

Implemented via https://github.com/alibaba/GraphScope/pull/4367

luoxiaojian (approved):

liulx20 (approved):

lnfjpt (approved):

PR #4410, opened 2025-01-09
feat(interactive): Introducing sharding mode to Interactive

zhanglei1949:
Introducing new configuration `sharding_mode` to Interactive. `sharding_mode` could be configured to `exclusive` or `cooperative`
- exclusive: One shard(shard_id = shard_num - 1) will be reserved for only processing admin request
- cooperative: All shard will processing both admin requests and query requests.

To specify this configuration, add in interactive_config.yaml
```yaml
http_service:
  sharding_mode: exclusive
```

By default, `sharding_mode` is configured to `exclusive` to forbid admin service from being blocked by long-run queries. User could switch the mode via a config file.

For the reason why this mode is introduced, see https://github.com/alibaba/GraphScope/discussions/4409

And also adding a CI and fix a bug.


LiSu (approved):

zhanglei1949:
The api-test failure will be resolved via #4432

longbinlai (approved):

PR #4415, opened 2025-01-10
fix(interactive): Support generating valid spring code with Interactive OpenAPI spec

zhanglei1949:
To support deserializing oneof field in python code


zhanglei1949:
Will be merged in #4411

PR #4426, opened 2025-01-14
fix(analytical): Fix Louvain algorithm's implementation

songqing:
## What do these changes do?

When I test Louvain with the [DBLP dataset](https://snap.stanford.edu/data/com-DBLP.html), I found the QUALITY is bigger than 1.0 (which is 1.13) in the first iteration, which is not correct, see [here](https://en.wikipedia.org/wiki/Louvain_method#:~:text=Modularity%20is%20a%20scale%20value,nodes%20of%20a%20given%20network.) QUALITY is in [-1, 1].
f1

Then I found the Louvain's implementation has some problems, as the Louvain  only supports undirected graph now, the current edge weight in each edge has calculated one more time as using both outgoing_edges() and incoming_edges(), fix it , we only need use outgoing_edges() in undirected graph.

### DIFF
--- a/analytical_engine/apps/pregel/louvain/louvain_app_base.h
+++ b/analytical_engine/apps/pregel/louvain/louvain_app_base.h
@@ -223,6 +223,7 @@ class LouvainAppBase
       auto actual_quality =
           ctx.compute_context().template get_aggregated_value<double>(
               actual_quality_aggregator);
+      assert(actual_quality <= 1.0);
       // after one pass if already decided halt, that means the pass yield no
       // changes, so we halt computation.
       if (current_super_step <= 14 ||
--- a/analytical_engine/apps/pregel/louvain/louvain_vertex.h
+++ b/analytical_engine/apps/pregel/louvain/louvain_vertex.h
@@ -68,7 +68,7 @@ class LouvainVertex : public PregelVertex<FRAG_T, VD_T, MD_T> {
 
   size_t edge_size() {
     if (!this->use_fake_edges()) {
-      return this->incoming_edges().Size() + this->outgoing_edges().Size();
+      return this->outgoing_edges().Size();
     } else {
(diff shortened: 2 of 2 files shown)


songqing:
@zhanglei1949 @siyuan0322 please help a review.

siyuan0322:
@zhanglei1949 plz check the error in PR-Check.

siyuan0322:
@songqing I don't think assert here is the best solution here since the whole program would crash... What about just log the error out as a reminder, since the actually value doesn't affect the clustering result, mainly focus on the trends.

songqing:
> @songqing Thanks for the fix! Just some little worries that I don't think assert here is the best solution here since the whole program would crash... What about just log the error out as a reminder, since the actually value doesn't affect the clustering result, mainly focus on the trends.

"Quality" can't be bigger than 1.0, if the the quality bigger than 1.0, then it means there is bug in the implementation, which needs to be fixed. Besides, there is other assert usage in the Louvain implementation I found.

siyuan0322:
> > @songqing Thanks for the fix! Just some little worries that I don't think assert here is the best solution here since the whole program would crash... What about just log the error out as a reminder, since the actually value doesn't affect the clustering result, mainly focus on the trends.
> 
> "Quality" can't be bigger than 1.0, if the the quality bigger than 1.0, then it means there is bug in the implementation, which needs to be fixed. Besides, there is other assert usage in the Louvain implementation I found.

Alright 😆

siyuan0322 (approved):

zhanglei1949:
> @zhanglei1949 plz check the error in PR-Check.

Seems to introduced by https://github.com/alibaba/GraphScope/commit/bb9a6032a3a2b88da0403c4607536cd9fa5c36d5. It is weird that the error didn't appeared in the commit that the change is introduced. 

I will fix it via another PR.

PR #4432, opened 2025-01-17
fix(interactive): Use static arrow when building interactive image

zhanglei1949:
In #4419, Interactive is changed to link to dynamic arrow lib but forgot to change the image building, where we need a static library for arrow.


longbinlai (approved):
LGTM

PR #4436, opened 2025-01-21
fix(interactive): Including primary keys when sinking vertex

### LINKED ISSUE #4433: [BUG] Primary key properties not returned.
When running cypher queries like `MATCH(n) return n`, the primary key property is not returned.

zhanglei1949:
Fix #4433

### DIFF
--- a/interactive_engine/compiler/src/main/java/com/alibaba/graphscope/common/ir/meta/fetcher/DynamicIrMetaFetcher.java
+++ b/interactive_engine/compiler/src/main/java/com/alibaba/graphscope/common/ir/meta/fetcher/DynamicIrMetaFetcher.java
@@ -81,7 +81,7 @@ public class DynamicIrMetaFetcher extends IrMetaFetcher implements AutoCloseable
             IrMeta meta = this.reader.readMeta();
             logger.debug(
                     "schema from remote: {}",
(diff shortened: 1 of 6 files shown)


interactive_engine/compiler/src/main/java/com/alibaba/graphscope/common/ir/meta/fetcher/DynamicIrMetaFetcher.java
```diff
@@ -81,7 +81,7 @@ private synchronized void syncMeta() {
             IrMeta meta = this.reader.readMeta();
             logger.debug(
                     "schema from remote: {}",
-                    (meta == null) ? null : meta.getSchema().getSchemaSpec(Type.IR_CORE_IN_JSON));
+                    (meta == null) ? null : meta.getSchema().getSchemaSpec(Type.FLEX_IN_JSON));
```
zhanglei1949, line 84: We need to use `FLEX_IN_JSON`, `IR_CORE_IN_JSON` doesn't support current data type system.

zhanglei1949: It seems that use FLEX_IN_JSON will cause other CI failure. This debug log is removed.

liulx20 (approved):

shirly121 (approved):

PR #4459, opened 2025-02-05
fix(interactive): Fix primary key check when fetching properties in runtime

### LINKED ISSUE #4460: [BUG] Merge the PropertyGetter the PKGetter for interactive runtime
As titled. 
Remove such fixed string comparsion: https://github.com/alibaba/GraphScope/blob/main/flex/engines/graph_db/runtime/utils/var.cc#L58

zhanglei1949:
In interactive, we store the primary key in a different way from common properties. To distinguish the primary key from other properties, we need to look up the schema at runtime.

Fix #4460


zhanglei1949:
There are still some issue not fixed via this PR, and marrked in https://github.com/alibaba/GraphScope/issues/4460

luoxiaojian (approved):

PR #4505, opened 2025-02-19
feat(interactive): `ImmutableCsr` support multiple properties

liulx20:
Fixes


zhanglei1949:
Maybe we need to change https://github.com/alibaba/GraphScope/blob/main/flex/storages/rt_mutable_graph/dual_csr.h#L103 
https://github.com/alibaba/GraphScope/blob/main/flex/storages/rt_mutable_graph/dual_csr.h#L429

to let it taking effect?

luoxiaojian (approved):

zhanglei1949 (approved):
LGTM

PR #4531, opened 2025-02-26
refactor(interactive): Enable routing to random shard when processing queries

zhanglei1949:
As titled.


zhanglei1949:
Closed since it will cause performance degradation.

PR #4538, opened 2025-02-27
refactor(interactive): Support complex graph schema

zhanglei1949:
Previously we only allow less than 256 labels for vertex and edges in Interactive. In this PR, we make `label_t` could be customized when building. And Introduce a test case to test interactive on a graph with 1000 vertex labels and 1000 edge labels.


flex/engines/graph_db/runtime/common/accessors.h
```diff
@@ -556,32 +557,23 @@ class MultiPropsEdgePropertyEdgeAccessor : public IAccessor {
   using elem_t = T;
 
   MultiPropsEdgePropertyEdgeAccessor(const GraphInterface& graph,
-                                     const std::string& name) {
+                                     const std::string& name)
+      : graph_(graph) {
     edge_label_num_ = graph.schema().edge_label_num();
     vertex_label_num_ = graph.schema().vertex_label_num();
-    indexs.resize(2 * vertex_label_num_ * vertex_label_num_ * edge_label_num_,
```
zhanglei1949, line 562: Seems no need to resize to 2* csr_num?

flex/engines/hqps_db/core/utils/hqps_utils.h
```diff
@@ -756,6 +756,13 @@ struct to_string_impl<uint8_t> {
   }
 };
 
+template <>
+struct to_string_impl<uint16_t> {
+  static inline std::string to_string(const uint16_t& empty) {
+    return std::to_string((int32_t) empty);
```
liulx20, line 762: should be `uint16_t` here

flex/storages/rt_mutable_graph/mutable_property_fragment.cc
```diff
@@ -29,17 +29,9 @@ MutablePropertyFragment::~MutablePropertyFragment() {
...
-                       dst_label * edge_label_num_ + e_label;
-        if (dual_csr_list_[index] != NULL) {
-          dual_csr_list_[index]->Resize(degree_list[src_label],
-                                        degree_list[dst_label]);
-          delete dual_csr_list_[index];
-        }
-      }
+  for (auto ptr : dual_csr_list_) {
+    if (ptr != NULL) {
+      delete ptr;
```
liulx20, line 34: call `resize` here to adjust to the actual size.

Issue #4043, opened 2024-07-15
[BUG] Can not return vertex/edge in procedure


zhanglei1949:
Exception in thread "main" java.lang.UnsupportedOperationException: can not convert RelDataType [Graph_Schema_Type(labels=[VertexLabel(Paper)], properties=[BIGINT id, CHAR(1) conference, CHAR(1) CCFRank, CHAR(1) CCFField, INTEGER year, CHAR(1) paper])] to GSDataTypeDesc
        at com.alibaba.graphscope.common.ir.meta.schema.GSDataTypeConvertor$Factory$2.convert(GSDataTypeConvertor.java:257)
        at com.alibaba.graphscope.common.ir.meta.schema.GSDataTypeConvertor$Factory$2.convert(GSDataTypeConvertor.java:103)
        at com.alibaba.graphscope.common.ir.meta.procedure.StoredProcedureMeta$Serializer.lambda$createProduceMetaMap$1(StoredProcedureMeta.java:191)
        at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
        at java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:948)
        at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
        at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
        at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913)
        at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:578)
        at com.alibaba.graphscope.common.ir.meta.procedure.StoredProcedureMeta$Serializer.createProduceMetaMap(StoredProcedureMeta.java:193)
        at com.alibaba.graphscope.common.ir.meta.procedure.StoredProcedureMeta$Serializer.perform(StoredProcedureMeta.java:153)
        at com.alibaba.graphscope.common.ir.tools.GraphPlanner.main(GraphPlanner.java:244)

PR #4556, opened 2025-03-10
feat(interactive): Support vertex/edge/path in FLEX data type

zhanglei1949:
As titled.


zhanglei1949:
Related issue: #4557 #4043

Issue #4581, opened 2025-03-25
[BUG] Insight failed on ldbc query is4


BingqingLyu:
**Describe the bug**

The query is as follows:

```
MATCH (m:POST|COMMENT {id: 1030792332314L})
RETURN
    CASE WHEN m.content = "" THEN m.imageFile
    ELSE m.content END as messageContent,
    m.creationDate as messageCreationDate
```

However, Insight outputs unexpected results. Fix it.

PR #4582, opened 2025-03-25
fix(interactive): Fix a bug in insight runtime when eval empty string

BingqingLyu:
## What do these changes do?

As titled, this bug occurs because an Object with an empty string is incorrectly treated as NULL, leading `NULL==NULL` to evaluate as false. However, the comparison `""==""` should be true. This PR addresses and fixes the issue.

## Related issue number

#4581


shirly121 (approved):

zhanglei1949 (approved):
LGTM

PR #4623, opened 2025-08-22
fix(analytical): Fix WCC's wrong result

songqing:
## What do these changes do?
When run this WCC algorithm with only one worker, the algorithm will only compute one iteration, so as to cause wrong result.
Fix by add ForceContinue() in PEval() when needed.


songqing:
@zhanglei1949 could you help merge this PR?

songqing:
> @songqing Could you run the failed checks?

@zhanglei1949 hi master, the failed UT before seems have nothing to do with my PR, and I rebase my PR now, please help trigger the check

